PR #9, opened 2017-05-02
Email verification

simonrozsival:


app/V1Module/presenters/EmailVerificationPresenter.php
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Resend the email for the current user to verify his/her email address.
+   * @POST
+   * @LoggedIn
+   */
+  public function actionResendVerificationEmail() {
+    $user = $this->getCurrentUser();
+    if (!$this->emailVerificationHelper->process($user)) {
+      throw new ForbiddenRequestException("You cannot request another verification email.");
```
Neloop: Process returns false if email cannot be sent...

PR #24, opened 2017-06-09
Exercise config form

Neloop:


SemaiCZE:
Seems ok, but let's wait until it's ready in the web-app.

PR #38, opened 2017-08-31
Exercise config validation

Teyras:


app/helpers/ExerciseConfig/Validation/ExerciseConfigValidator.php
```diff
@@ -59,49 +70,78 @@ private function checkEnvironments(ExerciseConfig $config, array $variablesTable
...
+        throw new ExerciseConfigException("Pipeline '$pipelineId' not found");
+      }
+
+      $pipeline = $this->loader->loadPipeline($pipelineEntity->getPipelineConfig()->getParsedPipeline());
+      $dataInBoxes = $pipeline->getDataInBoxes();
+      $inBoxNames = array_map(function (Box $box) { return $box->getName(); }, $dataInBoxes);
+      $variables = $pipelineVars->getVariablesTable();
+      $variableNames = array_map(function (Variable $box) { return $box->getName(); }, $variables->getAll());
+
+      if ($variableNames !== array_unique($variableNames)) {
```
Neloop: I believe this check is redundant, have a look at VariablesTable class... Also tiny detail: line 105 name of the variable $box is a bit misleading

app/helpers/ExerciseConfig/Validation/ExerciseConfigValidator.php
```diff
@@ -59,49 +70,78 @@ private function checkEnvironments(ExerciseConfig $config, array $variablesTable
...
+            $environment ?? "default"
+          ));
+        }
+
+        $inBoxNames = array_filter($inBoxNames, function (string $name) use ($variable) {
+          return $name !== $variable;
+        });
+      }
+
+      if (count($inBoxNames) > 0) {
```
Neloop: What if some variable is defined in environment config and not in exercise config? What happens here? Do I get and exception for not defining variable? If so, it is not the right behaviour.

app/helpers/ExerciseConfig/Validation/ExerciseLimitsValidator.php
```diff
@@ -35,14 +45,92 @@ public function __construct(Pipelines $pipelines) {
...
+
+      $environmentPipelines = [];
+
+      foreach ($pipelineNames as $name) {
+        $this->loadPipeline($name, $loadedPipelines);
+        $environmentPipelines[$name] = $loadedPipelines[$name];
+      }
 
-    // @todo: only identification of box is not enough in limits
+      if ($test === NULL) {
```
Neloop: Duplicit with line 53 or am I missing something?

app/helpers/ExerciseConfig/Validation/ExerciseLimitsValidator.php
```diff
@@ -35,14 +45,92 @@ public function __construct(Pipelines $pipelines) {
...
+
+          if ($box === NULL) {
+            throw new ExerciseConfigException(sprintf(
+              "Box with id %s does not exist in pipeline %s",
+              $boxId,
+              $pipelineId
+            ));
+          }
+
+          foreach ($limits as $limit => $value) {
```
Neloop: What is purpose of this, I do not really seem to understand why this is needed.

tests/ExerciseConfig/Validation/ExerciseConfigValidator.phpt
```diff
@@ -1,6 +1,6 @@
 <?php
 
-include '../../bootstrap.php';
+$container = include '../../bootstrap.php';
```
Neloop: Seems to be not used in here?

app/helpers/ExerciseConfig/Validation/ExerciseLimitsValidator.php
Teyras: Yep, good catch

app/helpers/ExerciseConfig/Validation/ExerciseConfigValidator.php
Teyras: This method is called per-environment - every pipeline is checked separately, using its variable table.

tests/ExerciseConfig/Validation/ExerciseConfigValidator.phpt
Teyras: check out line #227

app/helpers/ExerciseConfig/Validation/ExerciseLimitsValidator.php
Teyras: It makes sure that the user doesn't specify some nonsense limit, like "awesomeness: 9999"

tests/ExerciseConfig/Validation/ExerciseConfigValidator.phpt
Neloop: yes you are using it as a parameter for test class, but in the test class it is not used, or IntelliJ is trying to tell me it is not...

app/helpers/ExerciseConfig/Validation/ExerciseLimitsValidator.php
Neloop: And why is it problem? If user specifies it, it will not be saved in database, nor loaded into the structure. So this check wont see it...

Teyras: There is actually a Limits object and not an array, so the loop really is useless - I will remove it.

tests/ExerciseConfig/Validation/ExerciseConfigValidator.phpt
Teyras: My IntelliJ seems OK with it.

Neloop: Sorry poor choice of words. The constructor parameter is assigned to internal $this->container and this isntance container is not used anywhere else then in constructor. But this is only small detail.

PR #46, opened 2017-09-14
Default variables for runtime environments

Neloop:


fixtures/base/10_runtimes.neon
```diff
@@ -15,6 +20,11 @@ App\Model\Entity\RuntimeEnvironment:
             - "[pas]"
             - "GNU/Linux"
             - "Pascal compiled with FreePascal"
+            - '''
+              - name: "source_files"
+                type: "file[]"
```
Teyras: This should accept a single file (as long as we don't want to implement building complex pascal apps)

tests/Presenters/RuntimeEnvironmentsPresenter.phpt
```diff
@@ -26,7 +26,7 @@ class TestRuntimeEnvironmentsPresenter extends Tester\TestCase
   private $presenterPath = "V1:RuntimeEnvironments";
 
   /** @var RuntimeEnvironments */
-  protected $hardwareGroups;
+  protected $runtimeEnvironments;
```
Teyras: Context? Is that just a fix of bad naming?

Neloop: Just a fix which I ran into, while checking this presenter, somehow it worked without this variable.

fixtures/base/10_runtimes.neon
Neloop: Will be repaired in near future, together with appropriate box.

Teyras: OK

PR #48, opened 2017-09-17
Pipeline files

Neloop:


app/helpers/ExerciseFileStorage.php
```diff
@@ -31,13 +32,14 @@ public function __construct(FileServerProxy $fileServer) {
...
    * @param UploadedFile|FileUpload $file The file to be stored
-   * @param Exercise $exercise
-   * @return ExerciseFile|NULL If the operation is not successful, NULL is returned
+   * @param Exercise|null $exercise
+   * @param Pipeline|null $pipeline
+   * @return SupplementaryExerciseFile|NULL If the operation is not successful, NULL is returned
    * @internal param User $user User, who uploaded the file
    */
-  public function store(UploadedFile $file, Exercise $exercise) {
+  public function store(UploadedFile $file, Exercise $exercise = null, Pipeline $pipeline = null) {
```
Teyras: This is pretty ugly - can both arguments be null? It would be better to have two separate methods.

app/model/entity/SupplementaryExerciseFile.php
```diff
@@ -27,33 +27,60 @@ class ExerciseFile extends UploadedFile implements JsonSerializable
...
+    }
+
+    if ($pipeline) {
+      $this->pipelines->add($pipeline);
+      $pipeline->addSupplementaryEvaluationFile($this);
+    }
   }
 
-  public static function fromUploadedFile(UploadedFile $file, Exercise $exercise, string $hashName, string $fileServerPath) {
+  public static function fromUploadedFile(UploadedFile $file, ?Exercise $exercise, ?Pipeline $pipeline, string $hashName, string $fileServerPath) {
```
Teyras: Same as above. How about making a new entity for pipeline files?

PR #69, opened 2017-10-12
Superadmins can takover all users

Neloop:


Neloop:
For superadmin finding identifications of users is maybe even easier than emails. There is Users page which should contain these ids, therefore, ids were chosen.

PR #81, opened 2017-10-22
automatically lowercase uploaded file extensions and return error whe…

Teyras:
…n weird characters are used in uploaded file name

### DIFF
--- a/app/helpers/UploadedFileStorage.php
+++ b/app/helpers/UploadedFileStorage.php
@@ -2,16 +2,19 @@
 
 namespace App\Helpers;
 
+use App\Exceptions\InvalidArgumentException;
 use App\Model\Entity\UploadedFile;
 use App\Model\Entity\User;
 use App\Exceptions\UploadedFileException;
 use Nette;
 use Nette\Http\FileUpload;
+use Nette\Utils\Strings;
 
 /**
  * Stores uploaded files in a configured directory
  */
 class UploadedFileStorage extends Nette\Object {
+  public const FILENAME_PATTERN = '#^[a-z0-9\- _\.]+$#i';
 
   /** @var string Target directory, where the files will be stored */
   private $uploadDir;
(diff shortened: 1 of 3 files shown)


app/helpers/UploadedFileStorage.php
```diff
@@ -27,14 +30,26 @@ public function __construct(string $uploadDir) {
...
+   * @param User $user User who uploaded the file
    * @return UploadedFile|NULL If the operation is not successful, NULL is returned
+   * @throws InvalidArgumentException
    */
   public function store(FileUpload $file, User $user) {
     if (!$file->isOk()) {
       return NULL;
     }
 
+    if (!Strings::contains($file->getName(), ".") || Strings::endsWith($file->getName(), ".")) {
```
Neloop: What about files without extension? This can be case for additional files or supplementary files...

Teyras: Are you sure we want such files?

Teyras: Well... So far, we have like 18 of those and most of them are inputs/outputs.

Neloop: Yes this is exactly what I mean, inputs and outputs for example. I was under impression that files without extension are pretty normal on the UNIX systems and as an author of exercise I would not be happy to rename 20+ files just because of this.
What about @krulis-martin ?

krulis-martin: I agree witth @Neloop. Files without extension are quite common. Furthermore, files starting with dot are also common and should not be lowercased. My suggestion is to find last "." in the string and if it is not the first nor the last character in string, then lowercase everything after it.

Teyras: OK. Why should we allow a dot at the end of the file though?

krulis-martin: The question is, why should we prevent it. In general, it is better not to limit, what users can do.

Teyras: I understand that, but still, I can't see how it could be useful to have a file name ending with a dot.

krulis-martin: It would not be particularly useful. But given our "common" users, I would bet that at some point, someone will come with such a filename. And when he or she does, it will be yet another problem we will need to solve.
For the source files, I would not mind having more restrictive rules than for the names of uploaded files; however, I do not wish to restrict filenames for exercises more than absolutely necessary.

PR #83, opened 2017-10-23
Evaluate all submissions in-time, when broker sends them back

Neloop:


Neloop:
Will be on another PR, these features are not related.

PR #84, opened 2017-10-23
Exercise rights and intial state

krulis-martin:


krulis-martin:
Ignore Travis Errors for now, I still need to fix tests (as some assumptions were changed).

app/V1Module/security/Policies/ReferenceExerciseSolutionPermissionPolicy.php
```diff
@@ -0,0 +1,26 @@
...
+use App\Model\Entity\ReferenceExerciseSolution;
+use App\Security\Identity;
+
+class ReferenceExerciseSolutionPermissionPolicy implements IPermissionPolicy {
+  public function getAssociatedClass() {
+    return ReferenceExerciseSolution::class;
+  }
+
+  public function isAuthor(Identity $identity, ReferenceExerciseSolution $referenceExerciseSolution = null) {
+    if ($referenceExerciseSolution === null) {
```
Teyras: We use `NULL`, not `null`. The same goes for `true` and `false`.

krulis-martin: That is in violation of http://www.php-fig.org/psr/psr-2/

krulis-martin: null, true and false MUST be in lowercase

Teyras: Well, if you want to search for violations of PSR, there are plenty more :) IMO it's more important to be consistent... But this is a minor nitpick and it's possible we don't follow that rule so strictly anyway.

krulis-martin: Consistency is broken anyways. Lets shift this issue into another task (which will ensure consistency in one way or another).

Teyras: Agreed, have my approval :)

PR #86, opened 2017-10-24
WIP Recalculate points

Neloop:


Neloop:
Please do not merge until fully tested

app/helpers/Evaluation/EvaluationPointsLoader.php
```diff
@@ -0,0 +1,119 @@
...
+      // not a student submission
+      return;
+    }
+
+    // setup
+    $submission = $evaluation->getSubmission();
+    $maxPoints = $submission->getMaxPoints();
+
+    // calculate points from the score
+    $points = floor($evaluation->getScore() * $maxPoints);
```
krulis-martin: Would not round() be better?

app/V1Module/presenters/AssignmentsPresenter.php
```diff
@@ -182,16 +207,17 @@ public function actionUpdateDetail(string $id) {
...
-    // retrieve localizations and prepare some temp variables
-    $localizedTexts = $req->getPost("localizedTexts");
-    $localizations = [];
-
-    // localized texts cannot be empty
-    if (count($localizedTexts) == 0) {
-      throw new InvalidArgumentException("No entry for localized texts given.");
+    // if points were changed, go through all submissions and recalculate points
+    if ($oldFirstDeadlinePoints != $firstDeadlinePoints ||
+        $oldSecondDeadlinePoints != $secondDeadlinePoints) {
```
krulis-martin: Check also for changes in the deadlines themselves and change in pointsPercentualThreshold.

krulis-martin: On second thoughts, where are the deadlines applied? When assigning the points or when the best submission of a particular assignment is being determined to display final points? If they are not needed for the points recalculation, then there is no need to test their change...

app/helpers/Evaluation/EvaluationPointsLoader.php
Neloop: Not sure, this is original code, which I did not write.

app/V1Module/presenters/AssignmentsPresenter.php
Neloop: We need to check this too and recalculate even for deadlines... Points are assigned according to this information too.

app/helpers/Evaluation/EvaluationPointsLoader.php
SemaiCZE: Because 0.95 is almost 1, but it is not fully correct, so floor reflects this better.

krulis-martin: That may be true, but students receive the best grade even if their test/oral exam/... is not completely correct. I mean if you have like 20 tests (weighted evenly) and maxPoint == 3, it would be nice to assign 3 points even for 19 correct tests. Just my opinion.

SemaiCZE: Let's discuss this tomorrow, but we think that CodEx test are designed in a way that even one failed test means a penalty. If there is a `round()` function, it would be common that students will receive all of the points for suboptimal solutions and for example Mr. Jezek will complain instantly. We suppose, that each exercise have 10 - 20 tests and maximum of 10 points.

Teyras: I completely agree with @SemaiCZE here - if some tests are not necessary for a 100% score, it must be stated explicitly (which I think is not possible now, but that's another issue).

PR #87, opened 2017-10-24
Update reference solution endpoint to include flags whether current user can delete or evaluate

krulis-martin:


Neloop:
Not sure if we really settled on this approach... It somehow smells... Personally I would engage new permissions endpoint as soon as possible, including this.

Teyras:
I agree, adding a general permissions endpoint is really simple and cleaner than reimplementing this for every entity.

krulis-martin:
Well, since we have no permissions endpoint yet, and this needs to be done soon, the proposed solution was the effective and efficient way, how to implement it. And it is much better than re-implementing ACL verifications at client side.
Furthermore, nice implementation of the security endpoint will require much more work. I do not see the "really simple". Straightforward maybe, but it will take some doing.

Teyras:
An API endpoint that takes an array of structures that map 1:1 to ACL calls is really not complicated. However, I agree that using this efficiently from the web-app might be difficult.

krulis-martin:
But we do not want an API that resolves ACL calls. That would require to reimplement business logic at client side. You need also a mapping between which ACL functions call which endpoints to find out, whether API can call DELETE /exercise/<id>, for instance.
I have already enlisted a topic for tomorrows discussion regarding the endpoint. But before we have it, providing the canXXX flags in the response for each object is the best way to go in my opinion.

Teyras:
Well, we already have endpoints such as `/assignments/<id>/can-submit` that serve a similar purpose. Adding another way of checking permissions doesn't seem wise to me. I agree that in the web app, you should ideally only know the endpoints that you need to call and adding some privilege identifiers is not optimal.

For now, I propose to add a simple endpoint that checks if the user can update/delete a ref. solution and be done with it. Later, we can try to come up with a more general solution.

app/V1Module/presenters/ReferenceExerciseSolutionsPresenter.php
```diff
@@ -103,7 +103,14 @@ public function actionExercise(string $exerciseId) {
       throw new ForbiddenRequestException("You cannot access this exercise solutions");
     }
 
-    $this->sendSuccessResponse($exercise->getReferenceSolutions()->getValues());
+    $refSolutions = array_map(function($rs) use($exercise) {
+      $json = $rs->jsonSerialize();
+      $json['canDelete'] = $this->exerciseAcl->canDeleteReferenceSolution($exercise, $rs);
```
Teyras: I seem to have missed this yesterday. Does the `canDeleteReferenceSolution` method really need the exercise parameter? The exercise could be acquired with `$rs->getExercise()` in permission policy methods...

krulis-martin: Creating another endpoint, which would also be a temporary solution, does not make any sense. Especially, if it does not bring anything good (only a philosophical improvement on which we do not have consensus), means a lot of work (which will be dismissed later), and will increase the amount of requests (one per each button) not mentioning worsening UX. I think that the right course of action is either accept this solution as temporary, or go for the right solution (but it is needed really soon).

krulis-martin: Ugh, wrong thread. Yes, the `canDeleteReferenceSolution` can be simplified. Will do.

Neloop:
Well, if this should be temporary solution, why not to include it in web application directly as for every other role checks? If we do not know the final solution for this it seems logical...

krulis-martin:
Hints were moved into a separate decorator. However, it is still a temporary solution. I will rewrite the decorator to use security endpoint (once it is implemented). Furthermore, the flags are also temporary as we need to analyze, which permission hints are likely to be needed.

app/V1Module/presenters/base/BasePresenter.php
```diff
@@ -193,6 +201,9 @@ protected function sendSuccessResponse($payload, $code = IResponse::S200_OK) {
       $this->userActions->log($this->getAction(TRUE), $params, $code);
     }
 
+    if ($this->responseDecorator)
+      $payload = $this->responseDecorator->decorate($payload);
```
Neloop: Please use curly braces!

krulis-martin: Will do.

app/helpers/ResponseDecorator/IResponseDecorator.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\Helpers;
+
+/**
+ * Interface for decorators, which augment the presenter output just before it is passed on to JSON serialization.
+ */
+interface IResponseDecorator
```
Teyras: Do I understand correctly that there will always be only one decorator?

app/helpers/ResponseDecorator/PermissionsResponseDecorator.php
```diff
@@ -0,0 +1,46 @@
...
+    if (is_array($payload)) {
+      return array_map([$this, 'decorate'], $payload);
+    }
+
+    // This is just a proof of concept, which is used to demonstrate the functionality.
+    // The flags-permissions mapping will be in neon configuration file.
+    if ($payload instanceof ReferenceExerciseSolution) {
+      $json = $payload->jsonSerialize();
+      $json['canDelete'] = $this->exerciseAcl->canDeleteReferenceSolution($payload->getExercise(), $payload);
+      $json['canEvaluate'] = $this->exerciseAcl->canEvaluateReferenceSolution($payload->getExercise(), $payload);
```
Teyras: Personally, I would like it better if this was nested in another array (e.g. `$json["permissionHints"]`)

krulis-martin: I agree. However, this was a quick prototype and I wanted to test it without changing UI implementation first.

app/helpers/ResponseDecorator/IResponseDecorator.php
krulis-martin: Yes. However, If multiple decorators were needed, I will implement ComposedDecorator, which could have multiple decorators under it.

Teyras: Okay, the only thing that seems a little bit unclear is what is a good place for `$payload->jsonSerialize()` to happen. What if the decorator received an already serialized payload?

krulis-martin: Good point. I thought it is straightforward, but the composition will be a bit more difficult. I will give it some thought.

PR #91, opened 2017-10-29
User actions

Neloop:


app/helpers/UserActions.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    // construct content for logger
+    $contentArr = [];
+    $contentArr[] = $identity->getUserData()->getId();
+    $contentArr[] = (new DateTime)->getTimestamp();
+    $contentArr[] = $action;
+    $contentArr[] = Json::encode($params);
+    $contentArr[] = $code;
+    $contentArr[] = Json::encode($data);
```
krulis-martin: What about constructing the array as whole ... $contentArr = [ $identity-> ... ];

app/commands/DoctrineFixtures.php
```diff
@@ -115,6 +115,10 @@ protected function clearDatabase()
 
     foreach ($this->dbConnection->getSchemaManager()->listTables() as $table) {
       $tableName = $table->getName();
+      if ($tableName === "doctrine_migrations") {
+        // do not clear migrations table... it is crucial
+        continue;
+      }
```
Teyras: We'd probably be better off disabling table deletion completely...

app/helpers/UserActions.php
Neloop: Can do...

PR #92, opened 2017-10-29
check file content endpoint output for non-unicode characters and rem…

Teyras:
…ove them

### DIFF
--- a/app/V1Module/presenters/UploadedFilesPresenter.php
+++ b/app/V1Module/presenters/UploadedFilesPresenter.php
@@ -15,6 +15,7 @@ use App\Model\Repository\SupplementaryExerciseFiles;
 use App\Model\Repository\UploadedFiles;
 use App\Responses\GuzzleResponse;
 use App\Security\ACL\IUploadedFilePermissions;
+use ForceUTF8\Encoding;
 use Nette\Application\Responses\FileResponse;
 
 /**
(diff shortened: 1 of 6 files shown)


app/V1Module/presenters/UploadedFilesPresenter.php
```diff
@@ -99,7 +100,14 @@ public function actionContent(string $id) {
     if (!$this->uploadedFileAcl->canDownload($file)) {
       throw new ForbiddenRequestException("You are not allowed to access file '{$file->getId()}");
     }
-    $this->sendSuccessResponse($file->getContent());
+
+    $content = $file->getContent();
+    $fixedContent = Encoding::toUTF8($content);
```
krulis-martin: I would suggest to limit maximal size of the returned text (since it is kept in memory). If the file is larger than the limit, only it leading part (up to the limit) will be returned and "incomplete" flag will be added.

krulis-martin:
I have pushed a fix and implementation of my proposition. However, a minor task is still needed -- the file size limit needs to be moved to better place (e.g., configuration).

app/helpers/Config/UploadsConfig.php
```diff
@@ -10,12 +10,22 @@ class UploadsConfig extends Nette\Object {
...
+  protected $maxPreviewSize;
+
+  /**
    * UploadsConfig constructor.
    * @param array $config
    * @internal param string $removalThreshold
    */
   public function __construct(array $config) {
     $this->removalThreshold = Arrays::get($config, "removalThreshold", "1 day");
+    $this->maxPreviewSize = Arrays::get($config, "maxPreviewSize", 65536);
```
krulis-martin: Should not we place this also into neon config file (at least as an example of what can be configured)?

Teyras: Probably. However, thanks to us copying the config.neon.example for every deployment, our configs are kinda crowded. Maybe we should just document this stuff elsewhere.

krulis-martin: OK

PR #103, opened 2017-11-04
Fix permissions so a supervisor may see files of exercises he/she can assign

krulis-martin:


app/V1Module/security/Policies/UploadedFilePermissionPolicy.php
```diff
@@ -58,6 +58,30 @@ public function isAuthorOfSupplementaryFileExercises(Identity $identity, Uploade
...
+
+    $user = $identity->getUserData();
+    if ($user === null) {
+      return false;
+    }
+
+    foreach ($file->getExercises() as $exercise) {
+      foreach ($exercise->getGroups() as $group) {
+        if ($group->isAdminOrSupervisorOfSubgroup($user)) {
+          return true;  // The user can assign one of the corresponding exercises in hir group.
```
Teyras: Is `hir` a typo or a gender-neutral pronoun?

app/V1Module/security/Policies/UploadedFilePermissionPolicy.php
```diff
@@ -58,6 +58,30 @@ public function isAuthorOfSupplementaryFileExercises(Identity $identity, Uploade
     return FALSE;
   }
 
+
+  public function isInGroupUserSupervises(Identity $identity, UploadedFile $file)
```
Teyras: I understand that the name is long enough as is, but it should probably reflect the fact that this assertion only works for supplementary (test) files.

app/V1Module/security/Policies/UploadedFilePermissionPolicy.php
krulis-martin: The latter.

app/V1Module/security/Policies/UploadedFilePermissionPolicy.php
krulis-martin: Any suggestions?

Teyras: isSupplementaryInGroupUserSupervises? That's awfully long. Is it necessary for the method to only work on supplementary files? Don't additional files have a list of exercises too?

PR #109, opened 2017-11-16
add pipeline parameters

Teyras:


Neloop:
Relation between runtime environments and pipelines will be added later?
Pipelines which cannot be used in simple configuration will be determined by empty parameters array?

Teyras:
Thanks fo reading this! Adding a relationship between REs and pipelines is a good idea - I'll do that right away. There won't be pipelines with an empty parameters array - absence of an explicit parameter means we will send the default value.

Neloop:
Well ok, I can see sending default parameters now. But how are we gonna know difference between pipelines which can be used in simple form and which cannot? Because for pipelines which cannot be used in simple form is kinda nonsense having parameters which are no use in there.

Teyras:
By parameter values. In the simple config form, we'll just search for pipelines with parameters that say e.g. that the pipeline is a compilation pipeline or that it's an execution pipeline that consumes a file and produces something on the standard output. It's also possible to add a parameter that says "this is the default pipeline for this runtime environment" - what do you think of that? @krulis-martin

Neloop:
Ok, I can live with that. What about searching of parameters? Will be implemented later alongside with proper simple form endpoints?

Teyras:
IIRC, there will be no "simple form endpoints" - the searching will be done by the client, as well as dumping the user input into exercise config format (and loading it, too). There will be a field in the Exercise entity that determines the kind of editor (simple/full) that should be used for editation, but the rest will be handled by the client.

Neloop:
Yes of course, but you think that searching through all pipelines and finding of correct parameters should be implemented in client? In my opiniom a little helpfull endpoint which searches parameters is quite on point here.

Teyras:
Yeah, sure, I can implement an endpoint like that, if needed. @krulis-martin @SemaiCZE do you need that?

krulis-martin:
I think it would be better, if the client held all relevant pipelines for the simple config edit. I.e., we would like to download all pipelines marked as "pipelines for simple config" in one API request and then process (filter) them further at client side.

Neloop:
I am not sure, why process this on client... Because in exercise config, you're gonna exactly know which runtimes are needed, why download all and filter them if you can download only the necessary ones... ?

krulis-martin:
It would be nice to have pipelines for all environments. There are not too many of them and it would be nice to add/remove them without creating additional API calls.

Teyras:
But then you're going to filter the pipeline array anyway :)

PR #111, opened 2017-11-23
Custom judge

Neloop:


migrations/Version20171122181453.php
```diff
@@ -0,0 +1,165 @@
...
+
+      // save
+      $this->connection->executeQuery("UPDATE pipeline_config SET pipeline_config = :config WHERE id = :id",
+        ["config" => Yaml::dump($config), "id" => $pipelineConfig["id"]]);
+    }
+  }
+
+  /**
+   * To configurations using judges add judge arguments variable and custom
+   * judge file variable.
```
Teyras: Please, swap those two sentences. I'm not sure whether you can do this in English, but it doesn't read very well regardless of that.

fixtures/demo/15-pipelines.neon
```diff
@@ -106,6 +106,9 @@ App\Model\Entity\PipelineConfig:
                     judge-type:
                       type: string
                       value: ""
+                    args:
+                      type: string[]
+                      value: ""
```
Teyras: How about squashing these changes into the previous commit?

migrations/Version20171122181453.php
Neloop: This sentence is really badly build. I'll try to reformulate it.

PR #112, opened 2017-11-23
add a configuration type column to Exercise entity

Teyras:


app/helpers/ExerciseConfig/Compiler.php
```diff
@@ -18,6 +18,8 @@
  */
 class Compiler {
 
+  public const EXERCISE_CONFIG_TYPES = ["simpleExerciseConfig", "advancedExerciseConfig"];
```
Neloop: Whye here? It is not related to compilation, only to displaying data in web-app, I'd rather see this "list" in entity just like default parameters.

Teyras: Well, I thought about the entity too, but I don't think it belongs there either. There might be more compilation types in the future serviced by different compilers. Feel free to suggest a better place.

Neloop: Well, it is true that storing new configuration into ExerciseConfig entity might be one of the solutions. But it somehow smells. If you want to do this properly... then in the Compiler do a switch/if over config types...

Teyras: Okay, I'll do that.

Neloop: This of course means that you have to add type into Assignment entity...

Teyras: Why? I can access the exercise through the assignment entity, right?

Neloop: Yes, but what if type of the exercise will be changed in the meantime? Supervisor will use different type of confiugration, because he/she want to do exercise a bit more difficult...

Teyras: Oh yeah, Assignments can override ExerciseConfig... I need to think about this for a bit.

Neloop: well... cannot for now and I think we'd like to keep it that way. But in the database by manual override... yes...

Teyras: OK, copyiing config type to Assignment is probably the best option. Yuck.

app/helpers/ExerciseConfig/Compiler.php
```diff
@@ -52,6 +54,10 @@ public function compile(IExercise $exercise,
       RuntimeEnvironment $runtimeEnvironment,
       CompilationParams $params): JobConfig {
 
+    if (!$this->checkConfigurationType($exercise->getConfigurationType())) {
```
Neloop: Can you please call this as a static method and not instance one?

app/model/entity/ExerciseData.php
```diff
@@ -0,0 +1,170 @@
+<?php
```
Neloop: We might want to create new traits in the near future. Consider making new namespace, or at least new folder which will hold traits. It might become a little bit messy, what is actual entity and what is not...

Teyras: That's a good point, I'll move it.

PR #115, opened 2017-11-29
send an e-mail when submission cannot be sent to the broker

Teyras:


Teyras:
What exactly do you have in mind?

PR #117, opened 2017-11-30
New judge types

krulis-martin:


app/helpers/ExerciseConfig/Pipeline/Box/Boxes/JudgeBox.php
```diff
@@ -29,11 +29,7 @@ class JudgeBox extends Box
   public static $EXPECTED_OUTPUT_PORT_KEY = "expected-output";
   public static $DEFAULT_NAME = "ReCodEx Judge";
 
-  /* TYPES OF JUDGES */
-  public static $RECODEX_NORMAL_TYPE = "recodex-judge-normal";
-  public static $RECODEX_SHUFFLE_TYPE = "recodex-judge-shuffle";
-  public static $DIFF_TYPE = "diff";
-  public static $DIFF_BINARY = "/usr/bin/diff";
+  /** Translation of judge type to command and args. The first item is the default. */
```
Neloop: Why is this doc comment here?

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/JudgeBox.php
```diff
@@ -113,19 +109,32 @@ private function getJudgeBinaryAndArgs(): array {
...
     }
 
-    // judge type decision logic
-    if (empty($judgeType) || strtolower($judgeType) === self::$RECODEX_NORMAL_TYPE) {
-      return [ConfigParams::$JUDGES_DIR . self::$RECODEX_NORMAL_TYPE, []];
-    } else if (strtolower($judgeType) === self::$RECODEX_SHUFFLE_TYPE) {
-      return [ConfigParams::$JUDGES_DIR . self::$RECODEX_SHUFFLE_TYPE, []];
-    } else if (strtolower($judgeType) === self::$DIFF_TYPE) {
-      return [self::$DIFF_BINARY, []];
+    static $judgeTypes = null;
```
Teyras: Wouldn't it be better (readability-wise) to make this a class constant? It's not like lazy initialization saves us a considerable amount of time here...

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/JudgeBox.php
```diff
@@ -113,19 +109,32 @@ private function getJudgeBinaryAndArgs(): array {
...
+        'recodex-judge-shuffle-newline' => [ConfigParams::$JUDGES_DIR . 'recodex-judge-shuffle', ['-i','-n']], // judge ignoring order of tokens (which treats \n ...)
+        'diff' => ["/usr/bin/diff", []],                                                               // diff (binary-safe) judge
+      ];
     }
 
-    throw new ExerciseConfigException("Unknown judge type");
+    // judge type decision logic
+    if (empty($judgeType)) {
+      return reset($judgeTypes);
+    } else if (!empty($judgeTypes[$judgeType])) {
```
Teyras: I would probably use `array_key_exists` here - if the key is in the array, it is almost certain it will have a non-empty value.

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/JudgeBox.php
krulis-martin: It cannot be a constant nor class static member since the initialization uses non-constant expressions.

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/JudgeBox.php
krulis-martin: Because the author of this PR is an idiot, who cannot copy-paste stuff properly.

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/JudgeBox.php
krulis-martin: I would not for two reasons: array_key_exists is longer to write and is more error prone (the least thing I can do is to check that a nonempty value is really there).

Teyras: Okay, let's say it's a matter of personal taste... (array_key_exists is more explicit about what it does)

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/JudgeBox.php
Teyras: Which expressions are problematic? I have used array constants before (`const`). `ConfigParams::$JUDGES_DIR` could be made constant too.

krulis-martin: Variables (not even static ones like ConfigParams::$JUDGE_DIR) cannot be used in member variable (property) initialization nor in class constant initialization. Initialization expression must be constant (so it can be resolved at compile time).

PR #130, opened 2017-12-21
Make sure getCurrentUser() doesn't return null

Teyras:


app/V1Module/presenters/SubmitPresenter.php
```diff
@@ -260,8 +273,14 @@ private function finishSubmission(AssignmentSolution $solution, bool $isDebug =
         $solution->getSolution()->getFiles()->getValues(),
         $generatorResult->getJobConfig()
       );
-    } catch (\Exception $e) {
+
+      $this->entityManager->getConnection()->commit();
+    } catch (SubmissionFailedException $e) {
+      $this->entityManager->getConnection()->rollBack();
       $this->submissionFailed($submission, $e->getMessage());
```
Neloop: You are creating failure with the submission you just rollbacked.

Teyras: It looks like it miraculously worked anyway. Can you check my new implementation, please?

PR #132, opened 2017-12-24
Private public info

Neloop:


Neloop:
Can try...

app/V1Module/presenters/InstancesPresenter.php
```diff
@@ -178,30 +184,11 @@ public function actionGroups(string $id) {
...
-
-  /**
-   * Get a list of all public groups in an instance.
-   * @GET
-   * @param string $id An identifier of the instance
-   * @throws ForbiddenRequestException
-   */
-  public function actionPublicGroups(string $id) {
-    $instance = $this->instances->findOrThrow($id);
-    if (!$this->instanceAcl->canViewGroups($instance)) {
```
Teyras: Doesn't removing this mean that we can remove the `canViewGroups` privilege too?

Neloop: Not really sure... maybe... I'll look into it

Teyras: I see. Maybe stuff like [this](https://github.com/ReCodEx/api/blob/master/app/V1Module/presenters/InstancesPresenter.php#L179) should use GroupViewFactory too?

Neloop: You mean... like this: https://github.com/ReCodEx/api/blob/private-public-info/app/V1Module/presenters/InstancesPresenter.php#L191 ... ?

Teyras: Yeah, sorry for not checking the right branch.

PR #133, opened 2017-12-28
Solution view factory

krulis-martin:


app/model/repository/Comments.php
```diff
@@ -31,4 +32,57 @@ public function persistComment(Comment $comment) {
...
+    if (!$user) {
+      return 0; // anonymous users cannot see any comments
+    }
+
+    $qb = $this->comments->createQueryBuilder('tc')
+      ->select('COUNT(tc.id)')
+      ->where('tc.commentThread = :id');
+
+    if ($allVisible) {
+      $qb->andWhere('tc.user = :user OR tc.isPrivate = 0');
```
Teyras: It might be better to filter this based on ACL - the logic of what the isPrivate flag means should not be repeated here. This might lead e.g. to the comment being hidden from a superadmin. However, since this method only fetches the count of comments, I'm willing to turn my head.

app/model/repository/Comments.php
```diff
@@ -31,4 +32,57 @@ public function persistComment(Comment $comment) {
...
+    return $this->comments->matching(Criteria::create()
+      ->where(Criteria::expr()->andX(
+        Criteria::expr()->eq('commentThread', $thread),
+        Criteria::expr()->orX(
+          Criteria::expr()->eq('user', $user),
+          Criteria::expr()->eq('isPrivate', false)
+        )))
+      ->orderBy(["postedAt" => Criteria::DESC])
+      ->setMaxResults(1)
+    )->first();
```
Teyras: This could be written more succintly with DQL.

app/model/view/AssignmentSolutionViewFactory.php
```diff
@@ -61,6 +64,11 @@ public function getSolutionData(AssignmentSolution $solution) {
       "bonusPoints" => $solution->getBonusPoints(),
       "lastSubmission" => $solution->getLastSubmission() ? $solution->getLastSubmission()->getData($canViewDetails, $canViewValues) : null,
       "submissions" => $submissions,
+      "commentsStats" => $thread && $user ? [
+        "count" => $this->comments->getThreadCommentsCount($thread, $user),
+        "mineCount" => $this->comments->getThreadCommentsCount($thread, $user, false),
```
Teyras: The name of the column is not as clear as it could be. How about `authoredCount`?

app/model/repository/Comments.php
```diff
@@ -31,4 +32,57 @@ public function persistComment(Comment $comment) {
...
     $this->persist($thread);
   }
+
+  /**
+   * Get the number of visible comments for given user
+   * @param $threadId Id of the comments thread.
+   * @param $userId Id of the viewer or null (if nobody is logged in).
+   * @param bool $allVisible True = all coments visible for the user, false = comments made by the user.
+   */
+  public function getThreadCommentsCount(CommentThread $thread, ?User $user, bool $allVisible = true)
```
Teyras: The meaning of the boolean flag is not obvious in the usages of this method. How about making it private and implementing two methods with different names that call this one?

app/model/repository/Comments.php
```diff
@@ -31,4 +32,57 @@ public function persistComment(Comment $comment) {
...
+
+  /**
+   * Get the number of visible comments for given user
+   * @param $threadId Id of the comments thread.
+   * @param $userId Id of the viewer or null (if nobody is logged in).
+   */
+  public function getThreadLastComment(CommentThread $thread, ?User $user)
+  {
+    if (!$user) {
+      return null; // anonymous users cannot see any comments
```
Teyras: I would probably check this before calling the method and request a non-null $user, but feel free to dismiss this as a matter of personal taste.

app/model/repository/Comments.php
krulis-martin: Can do.

app/model/repository/Comments.php
krulis-martin: The whole point is NOT to fetch all entities and filter them using ACL, especially if we need the count.

app/model/repository/Comments.php
krulis-martin: It is also checked before call. This is just a paranoid preemptive measure.

app/model/repository/Comments.php
Teyras: I understand. But still, how about adding a comment that the method doesn't use the ACL?

app/model/view/AssignmentSolutionViewFactory.php
krulis-martin: OK

app/model/repository/Comments.php
Teyras: Then change the type in the signature and let the type system do the work for you :-)

krulis-martin: That was not the original intent, but very well, I do not mind.

app/model/view/AssignmentSolutionViewFactory.php
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Factory for group views which somehow do not fit into json serialization of
+ * entities.
+ */
+class AssignmentSolutionViewFactory {
+
+  /**
+   * @var IAssignmentSolutionPermissions
+   * @inject
```
Neloop: copy much?

PR #135, opened 2017-12-30
Multiple pascal input files

Neloop:


app/helpers/Config/ApiConfig.php
```diff
@@ -42,7 +52,10 @@ public function __construct(array $config) {
     $this->address = Arrays::get($config, ["address"]);
     $this->name = Arrays::get($config, ["name"]);
     $this->description = Arrays::get($config, ["description"]);
-    $this->version = Arrays::get($config, ["version"]);
+    $this->versionFormat = Arrays::get($config, ["versionFormat"], "{tag}");
+
+    // version is constructed from git version tag
+    $this->version = (new Directory(__DIR__))->get(new CustomFormatter($this->versionFormat));
```
Teyras: What happens if someone removes the `.git` directory? It might happen if we decided to ship RPM packages of the API...

Neloop: Yes, thats problematic... Do you have better solution in mind?

Teyras: I think it's OK (for now) if you just make sure that something like "UNKNOWN" will be filled in when reading the version from git fails.

PR #137, opened 2018-01-04
Local login

Neloop:


app/model/entity/Login.php
```diff
@@ -94,6 +94,14 @@ public function passwordsMatch($password) {
     return FALSE;
   }
 
+  /**
+   * Factory method.
```
Teyras: Does this docblock convey any useful information?

app/model/entity/User.php
```diff
@@ -316,4 +326,13 @@ public function getNameParts(): array {
       "degreesAfterName" => $this->degreesAfterName,
     ];
   }
+
+  /**
+   * True if user is have local login entity bound.
+   * @return bool
+   */
+  public function haveLocalAccounts(): bool {
```
Teyras: s/have/has

app/model/entity/User.php
```diff
@@ -316,4 +326,13 @@ public function getNameParts(): array {
       "degreesAfterName" => $this->degreesAfterName,
     ];
   }
+
+  /**
+   * True if user is have local login entity bound.
```
Teyras: Returns true if the user entity is associated with a local login entity.

app/model/entity/Login.php
Neloop: Just for satisfying PhpStorm...

Teyras: I can tolerate its existence, but you might want to consider turning that inspection off.

PR #140, opened 2018-01-11
Broken exercise flag

Teyras:


app/model/entity/Exercise.php
```diff
@@ -354,4 +388,12 @@ public function setLocked($value = TRUE) {
   public function clearExerciseLimits() {
     $this->exerciseLimits->clear();
   }
+
+  /**
+   * Get all runtime environments associated with the exercise
+   * @return Collection
+   */
+  public function getRuntimeEnvironments(): Collection {
```
Neloop: This should be in ExerciseData trait.

app/model/entity/Assignment.php
```diff
@@ -415,4 +415,12 @@ function ($key, RuntimeEnvironment $env) {
       ]
     ];
   }
+
+  /**
+   * Get all runtime environments associated with the exercise
+   * @return Collection
+   */
+  public function getRuntimeEnvironments(): Collection {
```
Neloop: This should be in ExerciseData trait.

app/helpers/ExerciseConfigChecker.php
```diff
@@ -0,0 +1,69 @@
+<?php
```
Neloop: Is there a reason why this file is not in ExerciseConfig subdir?

app/helpers/ExerciseConfigChecker.php
```diff
@@ -0,0 +1,69 @@
...
+   * Make up names of submitted files for a runtime environment.
+   * This is necessary because we do not have any real submissions yet, but the compiler needs their names.
+   * TODO when we implement a mechanism that ensures further constraints on submitted files, it must be reflected here
+   */
+  private function conjureSubmittedFiles(RuntimeEnvironment $environment) {
+    $extension = current($environment->getExtensionsList());
+    return ["main.{$extension}"];
+  }
+
+  public function check(Exercise $exercise) {
```
Neloop: Please comment this or whole class with appropriate description of what is done and how is exercise changed.

app/helpers/ExerciseConfigChecker.php
```diff
@@ -0,0 +1,69 @@
...
+    try {
+      $config = $this->loader->loadExerciseConfig($exercise->getExerciseConfig()->getParsedConfig());
+      $this->validator->validateExerciseConfig($exercise, $config);
+    } catch (ExerciseConfigException $exception) {
+      $exercise->setBroken(sprintf("Global exercise configuration is invalid: %s", $exception->getMessage()));
+      return;
+    }
+
+    /** @var RuntimeEnvironment $environment */
```
Neloop: After a year in php and PhpStorm... I am not really sure if I like this or not and if it is really necessary :-)

app/V1Module/presenters/ExerciseFilesPresenter.php
```diff
@@ -194,6 +200,10 @@ public function actionDeleteSupplementaryFile(string $id, string $fileId) {
     $exercise->updatedNow();
     $exercise->removeSupplementaryEvaluationFile($file);
     $this->exercises->flush();
+
+    $this->configChecker->check($exercise);
+    $this->exercises->flush();
```
Neloop: Is it really necessary flush exercises twice?

app/helpers/ExerciseConfigChecker.php
krulis-martin: Well, I am not fond of it either.

app/model/entity/Assignment.php
Teyras: Good call.

app/helpers/ExerciseConfigChecker.php
Teyras: Probably not, I'll move it there.

app/helpers/ExerciseConfigChecker.php
```diff
@@ -0,0 +1,69 @@
...
+
+use App\Exceptions\ExerciseConfigException;
+use App\Helpers\ExerciseConfig;
+use App\Helpers\ExerciseConfig\Compilation\CompilationParams;
+use App\Model\Entity\Exercise;
+use App\Model\Entity\RuntimeEnvironment;
+use Nette\SmartObject;
+
+class ExerciseConfigChecker {
+  use SmartObject;
```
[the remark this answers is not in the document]
Teyras: It gives you exceptions on access to undeclared properties.

app/helpers/ExerciseConfigChecker.php
Teyras: I'll gladly remove it when PHP starts supporting type declarations.

app/V1Module/presenters/ExerciseFilesPresenter.php
Teyras: I'm not 100% sure about this either. I think that setting the `broken` flag after validation should occur in a separate transaction so that it doesn't interfere with removing the file - that is a different operation. On the other hand, you could argue that if validation fails, the file should not be removed either. Any thoughts?

Neloop: Yeah, I though so (validator seeing changes).
As for the first thought... if we are setting broken flag, then in my opinion removing the file anyway is correct action to do.

app/helpers/ExerciseConfigChecker.php
Neloop: I know, I know... it wasn't meant to be complaint :-)

app/helpers/ExerciseConfigChecker.php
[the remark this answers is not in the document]
Neloop: Ok, I can live with that :-)

PR #141, opened 2018-01-13
Cleanup refactor

Neloop:


app/commands/cleanup/CleanupExerciseConfigs.php
```diff
@@ -0,0 +1,145 @@
...
+
+  /** @var EntityRepository */
+  private $assignments;
+
+  /** @var EntityManager */
+  private $entityManager;
+
+  public function __construct(EntityManager $entityManager) {
+    parent::__construct();
+    $this->exercises = $entityManager->getRepository(Exercise::class); // even deleted exercises has to be found
```
Teyras: Why don't you inject those classes directly (as constructor arguments)? `$em->getRepository()` doesn't guarantee you'll get an `EntityRepository` - you can configure the repository class in the `@ORM\Entity` annotation. Maybe adding methods that bypass the deletedAt filter would be a better solution.

readme.md
```diff
@@ -56,7 +56,7 @@ recommended way of ensuring this is using a crontab like this:
 
 ```
 04	00	*	*	*	php www/index.php notifications:assignment-deadlines "1 day"
-02	00	*	*	*	php www/index.php uploads:cleanup
+02	00	*	*	*	php www/index.php db:cleanup:uploads
```
Teyras: You could probably add the other commands here too.

Neloop: Well... it depends if we do want to do this regularly or not. In my opinion cleanups of configs and localized texts might by done once or twice a year, not every day or week...

app/commands/cleanup/CleanupExerciseConfigs.php
```diff
@@ -0,0 +1,145 @@
...
+    $deleteQuery->setParameter("ids", $usedLimits, Connection::PARAM_STR_ARRAY);
+    return $deleteQuery->execute();
+  }
+
+  protected function execute(InputInterface $input, OutputInterface $output) {
+    $now = new DateTime();
+    $limit = clone $now;
+    $limit->modify("-14 days");
+
+    $deletedEnv = $this->cleanupEnvironmentConfigs($limit);
```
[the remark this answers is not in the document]
Neloop: Ok, can do...

app/commands/cleanup/CleanupExerciseConfigs.php
Neloop: Ok, I'll add new method to base soft delete repository

PR #142, opened 2018-01-13
Allow group admins to edit group exercises

Teyras:


app/model/entity/Group.php
```diff
@@ -306,6 +306,24 @@ public function isAdminOrSupervisorOfSubgroup(User $user): bool {
     return false;
   }
 
+  /**
+   * User is admin of a group or any of its parents
+   * @param User $user
+   * @return bool
+   */
+  public function isAdminOfSuperGroup(User $user): bool {
+    $group = $this;
```
Neloop: There is no need to go through all parent groups again... there is `getAdminsIds()` method

app/config/permissions.neon
```diff
@@ -306,6 +306,23 @@ permissions:
       conditions:
           - exercise.isAuthor
 
+    - allow: true # Admin of the group of an exercise has most of the rights to the exercise
```
Neloop: Wouldn't it be better to add condition `exercise.isSuperGroupAdmin` to above `exercise.isAuthor` and separate `delete` actions into new permission rule?

Teyras: Conditions are joined with `&&`, so this wouldn't work. I still welcome suggestions for a good `||` syntax, though.

app/model/entity/Group.php
Teyras: That would work too. Do you think it's worth rewriting it?

app/config/permissions.neon
Neloop: I though something like `exercise.isAuthorOrSuperGroupAdmin` method... but not sure if it is a good idea

app/model/entity/Group.php
Neloop: Well... I think so, if I am not mistaken `isAdminOf()` method is using `getAdminsIds()` method so... you are doing searching which is not really necessary...

app/config/permissions.neon
Teyras: That sounds way too specific for my taste.

Neloop: Ok, fair enough...

app/model/entity/Group.php
Teyras: You mean `getAdminIds` doesn't do the searching under the hood?

Neloop: Look in the code please...

PR #143, opened 2018-01-14
Output from worker is currently divided into stdout and stderr

Neloop:


app/helpers/Evaluation/EvaluationResults/TaskResults/TaskResult.php
```diff
@@ -49,9 +51,12 @@ public function __construct(array $data) {
...
     if (isset($data[self::OUTPUT_KEY])) {
-      $this->output = $data[self::OUTPUT_KEY];
-    } else {
-      $this->output = "";
+      if (isset($data[self::OUTPUT_KEY][self::OUTPUT_STDOUT_KEY])) {
+        $this->output .= $data[self::OUTPUT_KEY][self::OUTPUT_STDOUT_KEY];
+      }
+      if (isset($data[self::OUTPUT_KEY][self::OUTPUT_STDERR_KEY])) {
+        $this->output .= $data[self::OUTPUT_KEY][self::OUTPUT_STDERR_KEY];
+      }
```
Teyras: We should probably make sure that there is at least a newline between stdout and stderr.

Neloop: Come on

PR #145, opened 2018-01-18
If the status is TO in sandbox results, return appropriate negative s…

Neloop:
…tatistics


app/helpers/Evaluation/EvaluationResults/Stats/Stats.php
```diff
@@ -195,4 +224,5 @@ public function wasKilled(): bool {
   public function __toString() {
     return Json::encode($this->data);
   }
+
```
Teyras: Remove this, please.

PR #146, opened 2018-01-18
Check for collision between remote files and submitted user files.

Neloop:


Neloop:
Well... no, but if I do this... then next question is if we can test all the boxes and their compilation and proper behaviour... of course we can, but is it really necessary?

Teyras:
Well, I don't think it's obvious that this code does what it's supposed to, so I would really like to see a test.

PR #147, opened 2018-01-19
Php js support

Neloop:


app/helpers/ExerciseConfig/Compilation/VariablesResolver.php
```diff
@@ -120,8 +120,11 @@ public function resolveForInputNodes(MergeTree $mergeTree, VariablesTable $exerc
       $outputPort = current($inputBox->getOutputPorts());
       $variableName = $outputPort->getVariable();
       $child = current($node->getChildren());
-      $inputPortName = array_search($node, $child->getParents());
+      if ($child === false) {
+        throw new ExerciseConfigException("Input port not found for variable {$variableName}");
```
Teyras: Shouldn't this be handled by the validator?

app/helpers/ExerciseConfig/Compilation/PipelinesMerger.php
```diff
@@ -142,7 +142,10 @@ private function mergeTrees(MergeTree $first, MergeTree $second): MergeTree {
     }
 
     // transform remaining output from first tree into classic array
-    $newFirstOutput = array_values($outVars);
+    $newFirstOutput = [];
+    foreach ($outVars as $outVar) {
```
Teyras: If only `array_map` was not the verbose beast it is in PHP...

app/helpers/ExerciseConfig/Compilation/VariablesResolver.php
Neloop: It definitely should, but was not... I ran into this error while testing PHP and better be sure that in there this cannot happen, I will create task for handling this in validator.

app/helpers/ExerciseConfig/Compilation/PipelinesMerger.php
Neloop: Well sorry, I am more like the classical foreach guy :-) but can change it if you want.

Teyras: That was meant to be a criticism of your code, just a little rant. Foreach is OK.

app/helpers/ExerciseConfig/Compilation/VariablesResolver.php
Teyras: OK, that sounds reasonable.

PR #148, opened 2018-01-22
Add isOrganizational and isArchived flags to groups

Teyras:


app/V1Module/presenters/GroupsPresenter.php
```diff
@@ -204,6 +224,60 @@ public function actionUpdateGroup(string $id) {
...
+  /**
+   * Set the 'isOrganizational' flag for a group
+   * @POST
+   * @Param(type="post", name="value", validation="bool", required=TRUE, description="The value of the flag")
+   * @param string $id An identifier of the updated group
+   * @throws ForbiddenRequestException
+   * @throws InvalidArgumentException
+   * @throws NotFoundException
+   */
+  public function actionSetOrganizational(string $id) {
```
krulis-martin: Not that I don't appreciate specialized endpoints, but I thought this will be set in the group edit form.

app/V1Module/presenters/UsersPresenter.php
```diff
@@ -313,14 +313,26 @@ public function actionGroups(string $id) {
...
-      "student" => $this->groupViewFactory->getGroups($user->getGroupsAsStudent()->getValues()),
+      "supervisor" => $this->groupViewFactory->getGroups($asSupervisor->getValues()),
+      "student" => $this->groupViewFactory->getGroups($asStudent->getValues()),
       "stats" => $user->getGroupsAsStudent()->map(
         function (Group $group) use ($user) {
           return $this->groupViewFactory->getStudentsStats($group, $user);
         }
-      )->getValues()
+      )->getValues(),
+      "archived" => [
```
krulis-martin: Someone, who uses ReCodEx for a long time, can get quite a lot (not always necessary) data here.

app/model/entity/Group.php
```diff
@@ -107,6 +112,65 @@ public function statsArePublic(): bool {
     return $this->publicStats;
   }
 
+  /**
+   * @ORM\Column(type="boolean")
+   */
+  protected $isArchived;
```
krulis-martin: Is not this a bit overkill? isArchived is equivalent to (archivationDate != NULL).

app/model/view/GroupViewFactory.php
```diff
@@ -164,8 +175,12 @@ public function getGroup(Group $group): array {
       "parentGroupId" => $group->getParentGroup() ? $group->getParentGroup()->getId() : null,
       "parentGroupsIds" => $group->getParentGroupsIds(),
       "childGroups" => [
-        "all" => $group->getChildGroupsIds(),
-        "public" => $group->getPublicChildGroupsIds()
+        "all" => $childGroups->map(function (Group $group) { return $group->getId(); })->getValues(),
+        "public" => $publicChildGroups->map(function (Group $group) { return $group->getId(); })->getValues()
+      ],
+      "archivedChildGroups" => [
```
krulis-martin: Similarly to the user entity situation, I believe that this is unnecessary and it will bite us in a few years. At present, we have 60 groups and the ReCodEx is not fully used right now. In the future, we can have easily 150 new groups each semester. That means after few years, we could easily get up to 1000 records being sent over. Archived data should be requested explicitly.

app/model/view/UserViewFactory.php
```diff
@@ -41,13 +41,25 @@ private function getUserData(User $user, bool $canViewPrivate) {
...
         "instanceId" => $user->getInstance()->getId(),
         "role" => $user->getRole(),
         "groups" => [
-          "studentOf" => $user->getGroupsAsStudent()->map(function (Group $group) { return $group->getId(); })->getValues(),
-          "supervisorOf" => $user->getGroupsAsSupervisor()->map(function (Group $group) { return $group->getId(); })->getValues()
+          "studentOf" => $studentOf->map(function (Group $group) { return $group->getId(); })->getValues(),
+          "supervisorOf" => $supervisorOf->map(function (Group $group) { return $group->getId(); })->getValues()
+        ],
+        "archivedGroups" => [
+          "studentOf" => $archivedStudentOf->map(function (Group $group) { return $group->getId(); })->getValues(),
```
krulis-martin: Same problem here.

app/V1Module/presenters/UsersPresenter.php
```diff
@@ -313,14 +313,49 @@ public function actionGroups(string $id) {
...
+    $asStudent = $user->getGroupsAsStudent()->filter(function (Group $group) {
+      return $group->isArchived();
+    });
+
+    $asSupervisor = $user->getGroupsAsSupervisor()->filter(function (Group $group) {
+      return $group->isArchived();
+    });
+
+    $this->sendSuccessResponse([
+      "supervisor" => $this->groupViewFactory->getGroups($asSupervisor->getValues()),
```
krulis-martin: I think a flag ignoreArchive is missing here in getGroups() call.

app/V1Module/presenters/GroupsPresenter.php
```diff
@@ -97,6 +99,20 @@ public function actionDefault() {
...
+   * @GET
+   * @throws ForbiddenRequestException
+   */
+  public function actionArchived() {
+    if (!$this->groupAcl->canViewAll()) {
+      throw new ForbiddenRequestException();
+    }
+
+    $groups = $this->groups->findArchived();
+    $this->sendSuccessResponse($this->groupViewFactory->getGroups($groups));
```
krulis-martin: I think a flag ignoreArchived is missing in getGroups() call.

PR #158, opened 2018-02-16
Submit oracle

Neloop:


app/helpers/ExerciseConfig/Helper.php
```diff
@@ -93,10 +134,119 @@ public function getVariablesForExercise(array $pipelines, VariablesTable $enviro
...
+          // we only seek for the local file variables, not the remote ones...
+          // although port might be of type file, variables assigned to it
+          // may have remote-file type in case the file should be downloaded
+          if (!$variable->isFile()) {
+            continue;
+          }
+
+          // determine if variable really defines only single wildcard expression
+          // and not an array of values
+          if ($variable->isValueArray()) {
```
krulis-martin: This may not be entirely correct. What if someone explicitly requires files with names 'implementation.hpp' and 'implementation.cpp'?

Neloop:
Pre-submit for reference solution will be added later, when reference solution will be redesigned in web-app.

PR #167, opened 2018-03-01
Allow disabling runtime environments in assignments

Teyras:


app/V1Module/presenters/AssignmentsPresenter.php
```diff
@@ -178,6 +188,13 @@ public function actionUpdateDetail(string $id) {
       throw new InvalidArgumentException("No entry for localized texts given.");
     }
 
+    if ($this->getHttpRequest()->getHeader("content-type") === "application/json") {
```
Neloop: Why is this necessary here?

app/model/entity/Assignment.php
```diff
@@ -14,6 +14,7 @@
 use DateTime;
 use Kdyby\Doctrine\MagicAccessors\MagicAccessors;
 use Gedmo\Mapping\Annotation as Gedmo;
+use Spatie\Regex\Helpers\Arr;
```
Neloop: Is this used somewhere?

app/model/entity/Assignment.php
```diff
@@ -407,13 +424,20 @@ function ($key, RuntimeEnvironment $env) {
             })
         ],
         "runtimeEnvironments" => [
-          "upToDate" => $this->getRuntimeEnvironments()->count() === $this->getExercise()->getRuntimeEnvironments()->count()
-            && $this->getRuntimeEnvironments()->forAll(function ($key, RuntimeEnvironment $env) {
+          "upToDate" => $this->runtimeEnvironments->count() === $this->getExercise()->getRuntimeEnvironments()->count()
```
Neloop: Well... question is, if synchronization should be advertised on change of environment which is disabled in assignment. This is rather unimportant for the assignment author. But not much of a big problem...

krulis-martin: Synchronization should not be advertised if the changes are in a disabled environment only. But I agree that it is not big issue.

app/V1Module/presenters/AssignmentsPresenter.php
Teyras: Because the meaning of empty `disabledRuntimeEnvironments` is ambiguous when multipart/form-data is used.

app/model/entity/Assignment.php
Teyras: Nope.

app/model/entity/Assignment.php
Teyras: If it's not a big issue, I would probably leave this as is for now. I feel like rephrasing the condition for enabled environments only won't be that easy.

app/V1Module/presenters/AssignmentsPresenter.php
Neloop: But this is quite ugly... Can you at least make some helper method here on BasePresenter or something like that?

app/model/entity/Assignment.php
Neloop: True would not be... Agree to postpone this

Teyras:
From my point of view the jedi are evil.

krulis-martin:
Then you are lost!

PR #169, opened 2018-03-08
Extract ACL checks and add security check endpoint.

Teyras:
Please review this carefully. Some methods could not have been split:

- create assignment
- upload supplementary files
- create/fork exercise
- add group
- add admin (add supervisor already has the ID in path)
- create pipeline
- fork pipeline
- sis create group
- bind group
- submit

### DIFF
--- a/app/V1Module/presenters/AssignmentSolutionsPresenter.php
+++ b/app/V1Module/presenters/AssignmentSolutionsPresenter.php
@@ -75,7 +75,7 @@ class AssignmentSolutionsPresenter extends BasePresenter {
   public function checkSolution(string $id) {
     $solution = $this->assignmentSolutions->findOrThrow($id);
     if (!$this->assignmentSolutionAcl->canViewDetail($solution)) {
-      throw new ForbiddenRequestException("You cannot change amount of bonus points for this submission");
+      throw new ForbiddenRequestException("You cannot view details of this solution");
     }
   }
 
--- a/app/V1Module/presenters/AssignmentSolutionsPresenter.php
+++ b/app/V1Module/presenters/AssignmentSolutionsPresenter.php
@@ -4,6 +4,7 @@ namespace App\V1Module\Presenters;
 
 use App\Exceptions\InternalServerErrorException;
 use App\Exceptions\NotFoundException;
+use App\Exceptions\NotReadyException;
 use App\Helpers\EvaluationLoadingHelper;
 use App\Helpers\FileServerProxy;
 use App\Model\Entity\AssignmentSolutionSubmission;
@@ -284,12 +285,6 @@ class AssignmentSolutionsPresenter extends BasePresenter {
     if (!$this->assignmentSolutionAcl->canDownloadResultArchive($submission->getAssignmentSolution())) {
       throw new ForbiddenRequestException("You cannot access result archive for this submission");
     }
-
-    $this->evaluationLoadingHelper->loadEvaluation($submission);
-
-    if (!$submission->hasEvaluation()) {
-      throw new ForbiddenRequestException("Submission is not evaluated yet");
-    }
   }
 
   /**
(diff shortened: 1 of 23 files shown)


app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -240,27 +263,31 @@ public function actionDownloadSolutionArchive(string $id) {
...
 
+  public function checkDownloadResultArchive(string $id) {
+    $submission = $this->assignmentSolutionSubmissions->findOrThrow($id);
+    if (!$this->assignmentSolutionAcl->canDownloadResultArchive($submission->getAssignmentSolution())) {
+      throw new ForbiddenRequestException("You cannot access result archive for this submission");
+    }
+
+    $this->evaluationLoadingHelper->loadEvaluation($submission);
+
+    if (!$submission->hasEvaluation()) {
```
Neloop: This in my opinion is not authorization check and should be in action method. User which have permissions to see button for downloading archive should see it even if it is not ready yet...

app/V1Module/presenters/GroupsPresenter.php
```diff
@@ -436,6 +468,18 @@ public function actionExercises(string $id) {
     $this->sendSuccessResponse($exercises);
   }
 
+  public function checkStats(string $id) {
+    $group = $this->groups->findOrThrow($id);
+
+    if (!$this->groupAcl->canViewStats($group)) {
+      $user = $this->getCurrentUser();
+
+      if (!($this->groupAcl->canViewStudentStats($group, $user) && $group->isStudentOf($user))) {
```
Neloop: Shouldn't this be negation of condition below on line 451?

app/V1Module/presenters/GroupsPresenter.php
```diff
@@ -512,6 +548,17 @@ public function actionAddStudent(string $id, string $userId) {
     if ($group->isOrganizational()) {
       throw new InvalidArgumentException("It is forbidden to add students to organizational groups");
```
Neloop: Again... not really sure if this should be in authorization check

app/V1Module/presenters/SecurityPresenter.php
```diff
@@ -0,0 +1,82 @@
...
+
+    try {
+      call_user_func_array([$presenter, $methodName], $arguments);
+    } catch (Exception $e) {
+      $result = false;
+    }
+
+    $this->sendSuccessResponse([
+      "result" => $result,
+      "certain" => true
```
Neloop: What exactly certain means?

app/V1Module/presenters/GroupsPresenter.php
Teyras: The method should check that the user either a) can view all statistics or b) is a student of the group and can view their own statistics. Is there any discrepancy between this and the code?

app/V1Module/presenters/GroupsPresenter.php
Teyras: I'm not certain either. However, I think that if the API call is likely to end with an InvalidArgumentException, we should not give the user false hopes in the check.

app/V1Module/presenters/AssignmentSolutionsPresenter.php
Teyras: Yeah, I'll probably give you this one.

app/V1Module/presenters/GroupsPresenter.php
Neloop: Well not convinced in either ways... independent opinion by @krulis-martin ?

app/V1Module/presenters/GroupsPresenter.php
Neloop: Sorrry... my bad... I did not see parenthesis

app/V1Module/presenters/GroupsPresenter.php
krulis-martin: In my opinion, the security endpoint check should follow the behavior of regular endpoint as closely as possible. I.e., if it is forbidden to add students to organizational group, then security should reflect this in authorization.
(Despite the name "security endpoint", I think the actual function is kind of preflight validation.)

Neloop: Ok, that is the thing, I am thinking of it as really just a security endpoint. But if it seems ok to you...

PR #170, opened 2018-03-10
Reference presubmit

Neloop:


app/V1Module/presenters/ReferenceExerciseSolutionsPresenter.php
```diff
@@ -407,7 +408,11 @@ private function evaluateReferenceSolution(
       $this->referenceSubmissions->flush();
     }
 
-    return [$evaluations, $errors];
+    return [
+      "referenceSolution" => $referenceSolution,
```
krulis-martin: I believe that the reference solution here will not hold the ID of the newly created submission in its `submissions[]`.

PR #173, opened 2018-03-15
Tweaks of disabling runtime environments

Teyras:


app/V1Module/presenters/AssignmentsPresenter.php
```diff
@@ -193,9 +193,12 @@ public function actionUpdateDetail(string $id) {
     }
 
     if ($this->isRequestJson()) {
-      $disabledRuntimeIds = $req->getPost("disabledRuntimeEnvironments");
-      $disabledRuntimes = array_map([$this->runtimeEnvironments, "findOrThrow"], $disabledRuntimeIds);
+      $disabledRuntimeIds = $req->getPost("disabledRuntimeEnvironmentIds");
```
Neloop: What if `$disableRuntimeIds` is null?

app/model/entity/Assignment.php
```diff
@@ -357,7 +357,9 @@ function ($key, RuntimeEnvironment $env) {
       "maxPointsBeforeFirstDeadline" => $this->maxPointsBeforeFirstDeadline,
       "maxPointsBeforeSecondDeadline" => $this->maxPointsBeforeSecondDeadline,
       "submissionsCountLimit" => $this->submissionsCountLimit,
-      "runtimeEnvironmentsIds" => $this->getRuntimeEnvironmentsIds(),
```
Neloop: just a thought, what about adding new item (`exerciseRuntimeEnvironmentIds`) and keep the old one?

app/V1Module/presenters/AssignmentsPresenter.php
```diff
@@ -193,10 +193,15 @@ public function actionUpdateDetail(string $id) {
...
+      if ($disabledRuntimeIds !== null) {
+        if (!is_array($disabledRuntimeIds)) {
```
Neloop: Instead of this you can use validation check in `@Param` annotation

Teyras: I'm not sure which one do you mean. Null or array validation?

Neloop: I'm refering to the line with array check... so I would say array validation

Teyras: Yeah, that could be done... I wasn't sure how validation works with optional parameters

PR #175, opened 2018-03-21
Implement endpoints for attaching and detaching exercises to groups

Neloop:


app/V1Module/router/RouterFactory.php
```diff
@@ -108,6 +108,8 @@ private static function createExercisesRoutes(string $prefix): RouteList {
     $router[] = new PostRoute("$prefix/<id>/fork", "Exercises:forkFrom");
     $router[] = new GetRoute("$prefix/<id>/pipelines", "Exercises:getPipelines");
     $router[] = new PostRoute("$prefix/<id>/hardware-groups", "Exercises:hardwareGroups");
+    $router[] = new PostRoute("$prefix/<id>/attach-group", "Exercises:attachGroup");
+    $router[] = new DeleteRoute("$prefix/<id>/detach-group", "Exercises:detachGroup");
```
Teyras: This could be a little more REST-ish - I would imagine something like PUT/DELETE `exercises/<exerciseId>/groups/<groupId>`. Another (perhaps more important) advantage would be that all the important information would be in the URL and not hidden in the request body (which is BTW illegal in DELETE requests), which is much better for checking permissions.

app/V1Module/security/ACL/IGroupPermissions.php
```diff
@@ -31,4 +31,6 @@ function canAssignExercise(Group $group, Exercise $exercise): bool;
   function canCreateExercise(Group $group): bool;
   function canViewPublicDetail(Group $group): bool;
   function canAddStudentToArchivedGroup($group, $user): bool;
+  function canAttachExercise(Group $group): bool;
+  function canDetachExercise(Group $group): bool;
```
Teyras: Isn't this the same as canAttachGroup? I would say these permissions should be symmetrical.

tests/Presenters/ExercisesPresenter.phpt
```diff
@@ -360,6 +360,55 @@ class TestExercisesPresenter extends Tester\TestCase
     Assert::equal("group1", $payload->getHardwareGroups()->first()->getId());
   }
 
+  public function testAttachGroup()
```
Teyras: Tests, yay! :+1:

app/V1Module/router/RouterFactory.php
Neloop: Good point, will change

app/V1Module/security/ACL/IGroupPermissions.php
Neloop: yes it is, we can consolidate this into one permission group, will do

app/V1Module/security/ACL/IGroupPermissions.php
```diff
@@ -31,6 +31,4 @@ function canAssignExercise(Group $group, Exercise $exercise): bool;
   function canCreateExercise(Group $group): bool;
   function canViewPublicDetail(Group $group): bool;
   function canAddStudentToArchivedGroup($group, $user): bool;
-  function canAttachExercise(Group $group): bool;
```
Teyras: :+1:

PR #180, opened 2018-03-29
Prevent some errors caused by broken exercises

Teyras:


app/helpers/Evaluation/IExercise.php
```diff
@@ -37,9 +37,9 @@ function getExerciseTests(): Collection;
 
   /**
    * Get configuration belonging to this exercise.
-   * @return ExerciseConfig
+   * @return ExerciseConfig|null
    */
-  function getExerciseConfig(): ExerciseConfig;
+  function getExerciseConfig(): ?ExerciseConfig;
```
Neloop: Well... ExerciseConfig should never be null, this hint is in my opinion pointless

app/helpers/ExerciseConfig/Compiler.php
```diff
@@ -65,8 +65,20 @@ public function compile(IExercise $exercise,
       throw new ExerciseConfigException("Submitted files contains two or more files with the same name.");
     }
 
+    if ($exercise->getExerciseConfig() === null) {
```
Neloop: The same as the above one...

app/V1Module/presenters/SubmitPresenter.php
```diff
@@ -268,11 +269,21 @@ private function finishSubmission(AssignmentSolution $solution, bool $isDebug =
...
+          $solution->getSolution()->getRuntimeEnvironment(),
+          $compilationParams);
+    } catch (ExerciseConfigException | JobConfigStorageException $e) {
+      $submission = new AssignmentSolutionSubmission($solution, "", $this->getCurrentUser());
+      $this->assignmentSubmissions->persist($submission, false);
+      $failure = SubmissionFailure::forSubmission(SubmissionFailure::TYPE_CONFIG_ERROR, $e->getMessage(), $submission);
+      $this->submissionFailures->persist($failure);
+      $this->failureHelper->report(FailureHelper::TYPE_API_ERROR, "Failed to generate job config for {$submission->getId()}");
+      throw new SubmissionFailedException($e->getMessage());
+    }
```
Neloop: I have to say, that I don't really like this, it is duplicit and easy to forget if something changes. Also if you do this you also have to solve this in reference submissions...

Neloop: What about using `submissionFailed` method?

app/helpers/Evaluation/IExercise.php
Teyras: And yet it happened in production not too long ago. That's why defensive programming is a thing.

app/helpers/ExerciseConfig/Compiler.php
Teyras: The same as the above one...

app/helpers/Evaluation/IExercise.php
Neloop: No... this did not happen... EnvironmentConfig was null, that is utterly different thing...

app/V1Module/presenters/SubmitPresenter.php
Teyras: We'd have to add a parameter or two, but it would be a better solution, yes. Reference submissions work a little differently, but I also checked there.

app/helpers/Evaluation/IExercise.php
Teyras: Hm, okay. Still, defensive programming :smile:

Neloop: Well... I am not gonna force you to change it, it is quite pointless... but defensive programming has to be taken seriously to some point... otherwise all we could do was checking this kind of errors...

app/V1Module/presenters/SubmitPresenter.php
Neloop: If it would be better, let's do that! Just a quick quesch, what is different in reference submissions?

Teyras: Reference submissions are typically added by people who know that something can (and oftend does) go wrong. The response also contains a field for errors that we can fill.

Neloop: KK... so lets change only failure helper method in normal submit...

PR #183, opened 2018-04-06
Set user role

Neloop:


app/config/permissions.neon
```diff
@@ -105,10 +105,17 @@ permissions:
           - update
           - remove
           - addSubgroup
-          - addSupervisor
           - removeSupervisor
       conditions: group.isAdmin
 
+    - allow: true
```
Teyras: This lets the superadmin make a student supervisor of a group. I'm not sure if that's okay...

app/V1Module/security/Policies/UserPermissionPolicy.php
```diff
@@ -31,4 +31,13 @@ public function isNotExternalAccount(Identity $identity, User $user): bool {
     return !$user->hasExternalAccounts();
   }
 
+  public function isSupervisor(Identity $identity, User $user) {
```
Teyras: An admin should also be a supervisor.

app/V1Module/presenters/UsersPresenter.php
```diff
@@ -430,4 +431,31 @@ public function actionExercises(string $id) {
     $this->sendSuccessResponse($user->getExercises()->getValues());
   }
 
+  public function checkSetRole(string $id) {
+    $user = $this->users->findOrThrow($id);
+    if (!$this->userAcl->canSetRole($user)) {
```
Teyras: Which users can set roles? A user with this privilege can make anybody a superadmin, which is pretty dangerous. Maybe we might want to have a non-admin user who can make supervisors...

app/config/permissions.neon
Neloop: Well, thats true, but according to this we should do the checks on supervisor role in set-role endpoint itself?

app/V1Module/security/Policies/UserPermissionPolicy.php
Neloop: Sorry, do not understand.

app/V1Module/presenters/UsersPresenter.php
Neloop: Like above, do you suggest that we should do the checks of this in the endpoint?

app/V1Module/security/ACLModule.php
```diff
@@ -6,19 +6,23 @@
...
     $this->authorizator = $authorizator;
+    $this->identity = $identity;
   }
 
   protected abstract function getResourceName();
 
   protected function check($action, $context): bool {
     return $this->authorizator->isAllowed(
-      $this->userStorage->getIdentity(),
+      $this->identity ?? $this->userStorage->getIdentity(),
```
krulis-martin: Would not it be better to set the identity (from arg or from userStorage) in the constructor?

PR #189, opened 2018-04-12
Mass endpoints

Neloop:


app/V1Module/presenters/UsersPresenter.php
```diff
@@ -64,10 +64,26 @@ public function checkDefault() {
...
     $users = $this->users->findAll();
-    $users = array_map(function (User $user) {
-      return $this->userViewFactory->getUser($user);
-    }, $users);
-    $this->sendSuccessResponse($users);
+    $this->sendSuccessResponse($this->userViewFactory->getUsers($users));
+  }
+
+  public function checkListByIds() {
+    if (!$this->userAcl->canViewList()) {
```
Teyras: Do we really need this privilege? And why does ExercisesPresenter use `canViewAll`?

Neloop: View all privilige on users have only superadmin, so we need different privilege which accesses this endpoint to normal users.
Good point on exercises though... view all privilege is accessible only to supervisors but this endpoint should be used by students too... will change

Teyras: I think the real question is whether we should keep `canViewAll` and friends or just return everything the user can see from the concerned endpoints. However, that's out of scope for this PR - feel free to merge.

PR #192, opened 2018-04-19
Data exercises

Neloop:


Neloop:
True dat... it's preparation for data exercises...

PR #193, opened 2018-04-19
Remove non-utf8 characters when previewing uploaded files

Teyras:


composer.json
```diff
@@ -5,10 +5,6 @@
 	"type": "project",
 	"license": ["MIT"],
 	"authors": [
-		{
-			"name": "Šimon Rozsíval",
-			"homepage": "https://github.com/simonrozsival"
-		},
```
krulis-martin: Poor Simon...

Teyras: A victim of collateral damage...

PR #196, opened 2018-04-19
Add useGravatar flag to user settings

Neloop:


migrations/Version20180419183337.php
```diff
@@ -0,0 +1,43 @@
...
+
+    $this->addSql('ALTER TABLE user_settings ADD use_gravatar TINYINT(1) NOT NULL COMMENT \'(DC2Type:boolean)\'');
+  }
+
+  /**
+   * Set newly added use gravatar flag to true.
+   * @param Schema $schema
+   */
+  public function postUp(Schema $schema) {
+    $this->connection->executeQuery("UPDATE user_settings SET use_gravatar = TRUE");
```
Teyras: Is TRUE actually allowed here?

Neloop: Well it worked on my local database (expectably) and even on our dev environment... So I would think so.

krulis-martin: It is far better, to use specific functions for specific purposes. Instead of `executeQuery`, you should have used `update()`. https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#update

PR #197, opened 2018-04-20
Do not allow create top-level exercises and detaching last group.

Neloop:


app/V1Module/presenters/ExercisesPresenter.php
```diff
@@ -442,6 +446,10 @@ public function checkDetachGroup(string $id, string $groupId) {
     if (!$this->exerciseAcl->canDetachGroup($exercise, $group)) {
       throw new ForbiddenRequestException("You are not allowed to detach the group to the exercise");
     }
+
+    if ($exercise->getGroups()->count() === 1) {
+      throw new ForbiddenRequestException("You cannot detach last group from exercise");
```
Teyras: This is not a forbidden request, it's more of a logic exception...

PR #199, opened 2018-04-23
Perf opt - check thoroughly!

Neloop:


app/model/repository/AssignmentSolutions.php
```diff
@@ -63,30 +83,42 @@ public function findValidSolutions(Assignment $assignment, User $user) {
   }
 
   /**
-   * @param Assignment $assignment
+   * Find best solutions of given assignments for user.
+   * @param Assignment[] $assignments
    * @param User $user
-   * @return AssignmentSolution|null
+   * @return Pair[] list of pairs where first is assignment and second solution, indexed by assignment id
```
Teyras: An (associative array indexed by assignment id) of pairs where the first item is an assignment entity and the second is a solution entity (the best one for the assignment)

PR #200, opened 2018-04-25
Updating exercise config presenter ...

krulis-martin:
... so that update environment config returns the updated data instead of OK message.


app/V1Module/presenters/ExercisesConfigPresenter.php
```diff
@@ -129,6 +129,23 @@ public function checkGetEnvironmentConfigs(string $id) {
     }
   }
 
+
+  /**
+   * Local micro-view factory, which constructs a response with environment configs of given exercise.
+   */
+  private function getEnvironmentConfigs(Exercise $exercise)
+  {
```
Neloop: Bad formatting

PR #201, opened 2018-04-26
Group views filtering

Teyras:
Breaking changes:
- sisCode disappeared from Group entity (replaced with private.bindings.sis)
- assignments array was flattened in Group entity
- childGroups array was flattened in Group entity

### DIFF
--- a/app/helpers/GroupBindings/GroupBindingAccessor.php
+++ b/app/helpers/GroupBindings/GroupBindingAccessor.php
@@ -17,7 +17,11 @@ class GroupBindingAccessor {
   public function getBindingsForGroup(Group $group) {
     $result = [];
     foreach ($this->providers as $provider) {
-      $result[$provider->getGroupBindingIdentifier()] = $provider->findGroupBindings($group);
+      $result[$provider->getGroupBindingIdentifier()] = [];
+
+      foreach ($provider->findGroupBindings($group) as $binding) {
+        $result[$provider->getGroupBindingIdentifier()][] = (string) $binding;
+      }
     }
     return $result;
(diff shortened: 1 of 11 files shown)


app/helpers/GroupBindings/GroupBindingAccessor.php
```diff
@@ -0,0 +1,24 @@
...
+  private $providers = [];
+
+  public function addProvider(IGroupBindingProvider $provider) {
+    $this->providers[] = $provider;
+  }
+
+  public function getBindingsForGroup(Group $group) {
+    $result = [];
+    foreach ($this->providers as $provider) {
+      $result[$provider->getGroupBindingIdentifier()] = $provider->findGroupBindings($group);
```
krulis-martin: This is not working correctly with serialization. The GroupBinding objects are serialized as empty objects, but we need their codes.

app/model/view/GroupViewFactory.php
```diff
@@ -31,9 +33,19 @@ class GroupViewFactory {
...
-  public function __construct(AssignmentSolutions $assignmentSolutions, IGroupPermissions $groupAcl) {
+  /**
+   * @var IAssignmentPermissions
+   */
+  private $assignmentAcl;
+
+  /** @var GroupBindingAccessor */
+  private $bindings;
+
+  public function __construct(AssignmentSolutions $assignmentSolutions, IGroupPermissions $groupAcl, IAssignmentPermissions $assignmentAcl, GroupBindingAccessor $bindings) {
```
krulis-martin: Line too long

PR #205, opened 2018-05-03
Add localized assignment text entity

Teyras:


app/V1Module/presenters/AssignmentsPresenter.php
```diff
@@ -131,7 +138,7 @@ public function checkDefault() {
    */
   public function actionDefault() {
     $assignments = $this->assignments->findAll();
-    $this->sendSuccessResponse($assignments);
+    $this->sendSuccessResponse(array_map([$this->assignmentViewFactory, "getAssignment"], $assignments));
```
Neloop: Not really fan of this...

app/model/view/AssignmentViewFactory.php
```diff
@@ -0,0 +1,89 @@
...
+  public function getAssignment(Assignment $assignment) {
+    return [
+      "id" => $assignment->getId(),
+      "version" => $assignment->getVersion(),
+      "isPublic" => $assignment->isPublic(),
+      "createdAt" => $assignment->getCreatedAt()->getTimestamp(),
+      "updatedAt" => $assignment->getUpdatedAt()->getTimestamp(),
+      "localizedTexts" => $assignment->getLocalizedTexts()->map(function (LocalizedExercise $text) use ($assignment) {
+        $data = $text->jsonSerialize();
+        if (!$this->assignmentAcl->canViewPrivateData($assignment)) {
```
Neloop: In my opinion `canViewPrivateData` (aka. private data) is really misleading here. As it is not private in the sense only visible to the author of assignment, but only visible to the supervisors. A better naming would be on point here.

app/V1Module/presenters/AssignmentsPresenter.php
```diff
@@ -170,6 +171,7 @@ public function checkUpdateDetail(string $id) {
    * @Param(type="post", name="version", validation="numericint", description="Version of the edited exercise")
    * @Param(type="post", name="isPublic", validation="bool", description="Is the assignment ready to be displayed to students?")
    * @Param(type="post", name="localizedTexts", validation="array", description="A description of the assignment")
+   * @Param(type="post", name="localizedAssignments", validation="array", description="Public localized hints for the assignment", required=false)
```
Neloop: Should be required and there should be tests of it ;-)

app/V1Module/presenters/AssignmentsPresenter.php
Teyras: What exactly?

Neloop: Defining the callback as the array... but its a php thing and not a deal breaker

Teyras: Yeah, that's not my jam either.  We could use `Nette\Utils\Callback` here if you want to.

krulis-martin: Unfortunately, only way to avoid delegate (meaning method+object as callback) construction as a two-item array is to create a wrapping anonymous function. Possibly more comprehensible, but more verbose.

PR #208, opened 2018-05-05
Debug stdout stderr

Neloop:


app/helpers/ExerciseConfig/Pipeline/Box/Boxes/base/CompilationBox.php
```diff
@@ -45,8 +47,15 @@ protected function compileBaseTask(CompilationParams $params): Task {
...
     $task->setFatalFailure(true);
 
-    $task->setSandboxConfig((new SandboxConfig)
-      ->setName(LinuxSandbox::$ISOLATE)->setOutput(true));
+    $sandbox = (new SandboxConfig)->setName(LinuxSandbox::$ISOLATE)->setOutput(true);
+    $task->setSandboxConfig($sandbox);
+
+    if ($params->isDebug()) {
+      $sandbox->setStderrToStdout(true);
+      $stdoutRandom = Random::generate(20) . ".compilation.out";
```
krulis-martin: A small favor to ask ...could this be "compilation.randomnum.out".

Neloop: Well... it could be... The question is if it should be...

Neloop: find works with either ways ¯\_(ツ)_/¯
but ok... tonite we will feast on new implementation of your proposal

krulis-martin: True. But when you are used to see files in TotalCommander/mc/whateverfilemanageryouuse, you might be used to see them (and look them up) ordered by name. In such case, stable prefix is a benefit.

PR #215, opened 2018-05-17
Mail presenter

Neloop:


app/helpers/Emails/EmailHelper/EmailHelper.php
```diff
@@ -40,20 +46,28 @@ public function __construct(IMailer $mailer, array $params) {
...
+  public function send(?string $from, array $to, string $subject, string $text, array $bcc = []) {
+    $subject = $this->subjectPrefix . $subject;
+    if ($from === null) {
+      // if from email is not provided use the default one
+      $from = $this->from;
+    }
+
     $latte = new Latte\Engine;
-    $latte->setTempDirectory(__DIR__ . "/../../../temp");
+    $latte->setTempDirectory(__DIR__ . "/../../../../../temp");
```
Teyras: This is nasty. I suggest we make the temp directory a configuration parameter and use `%tempDir%` in `config.neon`.

app/helpers/Emails/EmailHelper/EmailHelper.php
```diff
@@ -40,20 +46,28 @@ public function __construct(IMailer $mailer, array $params) {
...
-   * @param string $from Sender of the email
+   * @param string|null $from Sender of the email
    * @param array $to Receivers of the email
    * @param string $subject Subject of the email
    * @param string $text Text of the message
    * @param array $bcc Blind copy receivers
    * @return bool If sending was successful or not
    */
-  public function send(string $from, array $to, string $subject, string $text, array $bcc = []) {
+  public function send(?string $from, array $to, string $subject, string $text, array $bcc = []) {
```
Teyras: Couldn't we make a separate method to avoid calling `->send(null, "someone@example.com", ...)`? I mean something like `sendFromDefaultAddress`.

app/config/config.neon
```diff
@@ -89,43 +89,44 @@ parameters:
...
-      failureResolvedPrefix: "ReCodEx Submission Failure resolved - "
+      failureResolvedPrefix: "Submission Failure Resolved - "
 
   forgottenPassword:  # user messages for changing passwords
     redirectUrl: "https://recodex.mff.cuni.cz/en/forgotten-password/change"  # URL of web application where the password can be changed
     tokenExpiration: 600  # expiration timeout of temporary token (in seconds), 10 minutes
     emails:  # additional info for sending mails, from is source address and subjectPrefix is prefix of mail subject
       from: %emails.from%
-      subjectPrefix: "ReCodEx Forgotten Password Request - "
+      subjectPrefix: "Forgotten Password Request - "
```
Teyras: Since you're already touching this - this subject doesn't really make sense. How about "Password reset/recovery request"?

app/config/config.neon
```diff
@@ -89,43 +89,44 @@ parameters:
...
     emails:
       from: %emails.from%
-      submissionEvaluatedPrefix: "ReCodEx Submission Evaluated Notification - "
+      submissionEvaluatedPrefix: "Submission Evaluated - "
 
   failureResolutionNotifications:
     emails:
       from: %emails.from%
-      failureResolvedPrefix: "ReCodEx Submission Failure resolved - "
+      failureResolvedPrefix: "Submission Failure Resolved - "
```
Teyras: If I'm not mistaken, these strings are already hardcoded as defaults in the email helpers. Shouldn't we move this to the example config or remove it completely?

app/V1Module/presenters/EmailsPresenter.php
```diff
@@ -100,4 +115,34 @@ public function actionSendToRegularUsers() {
...
+
+  /**
+   * Sends an email with provided subject and message to all members of given group.
+   * @POST
+   * @param string $groupId
+   * @Param(type="post", name="subject", validation="string:1..", description="Subject for the soon to be sent email")
+   * @Param(type="post", name="message", validation="string:1..", description="Message which will be sent, can be html code")
+   * @throws NotFoundException
+   */
+  public function actionSendToGroupMembers(string $groupId) {
```
Teyras: Do we really want to send the message to supervisors and admins too? I could imagine a flag here.

app/helpers/Emails/EmailHelper/EmailHelper.php
Neloop: ¯\_(ツ)_/¯ This was there before, but if you'd like, I can make it disappear

app/helpers/Emails/EmailHelper/EmailHelper.php
Neloop: We definitely can

app/config/config.neon
Neloop: Ok, I was not really reading and reviewing all of these

app/config/config.neon
Neloop: Well, let me just say that I am not really fan of the `Arrays::get("key", "default")` in code.
Because of its duplication :-) ... In my opinion all configurable items should be defined in global configuration (`config.neon`) and the ones which really needs to change on every deployment in `config.local.neon`.
Why in global configuration rather then in code? Because its more readable to have all configuration items somewhere on the one place, so the administrator can decide what to override and what not. If these items would not be there as you suggest, nobody would know that it can be changed, because only few people will actually read the code.

app/V1Module/presenters/EmailsPresenter.php
Neloop: Yes, that is maybe better solution to have one action for a group and have switches for the particular user membership groups.

krulis-martin: I would like to request one more flag here: It would be nice to have "I want a copy of that email" option. It may be useful when I want a copy, but I do not want to notify all admins or when I am a superadmin managing group that is not mine.

Neloop: Fair enough... Can do

PR #216, opened 2018-05-17
Add permission hints to certain entity views

Teyras:


app/helpers/PermissionHints.php
```diff
@@ -0,0 +1,36 @@
...
+        continue;
+      }
+
+      $parameter = $method->getParameters()[0];
+      $className = $parameter->getClass() ? $parameter->getClass()->getName() : null;
+      if ($className !== null && $subject instanceof $className) {
+        yield lcfirst(substr($method->getName(), 3)) => $method->invoke($aclModule, $subject);
+      }
+    }
+  }
```
krulis-martin: I must say I am not happy about this solution. I do not mind using reflection, but imagine when the same type of entity (e.g., exercise) is decorated in the view using this hints generator many times. It is rather crude to iterate over an ACL class every time hints need to be added to an entity. I would suggest introducing some caching or adjusting the interface so that a generator (which already knows the right ACL methods) is constructed for the view, so the view may use it more efficiently to add hints to multiple entities.

app/helpers/ResponseDecorator/PermissionsResponseDecorator.php
```diff
@@ -1,49 +0,0 @@
...
-    }
-
-    // This is just a proof of concept, which is used to demonstrate the functionality.
-    // The flags-permissions mapping will be in neon configuration file.
-    if ($payload instanceof ReferenceExerciseSolution) {
-      $json = $payload->jsonSerialize();
-      $json['permissionHints'] = [
-        'delete' => $this->exerciseAcl->canDeleteReferenceSolution($payload->getExercise(), $payload),
-        'evaluate' => $this->exerciseAcl->canEvaluateReferenceSolution($payload->getExercise(), $payload),
-      ];
```
krulis-martin: I do not mind removing this obsolete decorator in this PR, but a replacement must be created immediately, so we need only one patch in the UI. Please, implement ACL hints for Reference Solutions as well.

app/helpers/PermissionHints.php
```diff
@@ -0,0 +1,36 @@
...
+use Nette\Reflection\Method;
+use Nette\StaticClassException;
+use Nette\Utils\Strings;
+
+class PermissionHints {
+  public function __construct() {
+    throw new StaticClassException();
+  }
+
+  public static function generate($aclModule, $subject) {
```
krulis-martin: This method is in dire need of doc comment. I must have read it twice and slowly to understand it completely.

PR #222, opened 2018-06-07
Temporary custom validation for overridenPoints which needs to be cha…

Neloop:
…nged in the future

### DIFF
--- a/app/V1Module/presenters/AssignmentSolutionsPresenter.php
+++ b/app/V1Module/presenters/AssignmentSolutionsPresenter.php
@@ -2,11 +2,14 @@
 
 namespace App\V1Module\Presenters;
 
+use App\Exceptions\BadRequestException;
 use App\Exceptions\InternalServerErrorException;
+use App\Exceptions\InvalidArgumentException;
 use App\Exceptions\NotFoundException;
 use App\Exceptions\NotReadyException;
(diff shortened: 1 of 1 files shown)


app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -189,17 +192,28 @@ public function checkSetBonusPoints(string $id) {
...
     $newBonusPoints = $this->getRequest()->getPost("bonusPoints");
     $overriddenPoints = $this->getRequest()->getPost("overriddenPoints");
     $solution = $this->assignmentSolutions->findOrThrow($id);
 
     $solution->setBonusPoints($newBonusPoints);
-    $solution->setOverriddenPoints($overriddenPoints === "" ? null : $overriddenPoints);
+
+    // TODO: validations 'null|numericint' for overridenPoints cannot be used, because null is converted to empty string,
+    // TODO: which immediately breaks stated validation... in the future, this behaviour has to change
+    // TODO: lucky third TODO
```
krulis-martin: Actually, you have four TODOs there, which breaks the luck...

Neloop: Darn it

PR #224, opened 2018-07-17
Pagination improvements

krulis-martin:


app/model/repository/Users.php
```diff
@@ -19,6 +21,77 @@ public function getByEmail(string $email): ?User {
...
+    // Set filters ...
+    if ($pagination->hasFilter("search")) {
+      $search = trim($pagination->getFilter("search"));
+      if (!$search) {
+        throw new InvalidArgumentException("filter", "search query value is empty");
+      }
+
+      $expr = $qb->expr()->orX();
+      foreach (["u.firstName", "u.lastName"] as $column) {
+        $expr->add($qb->expr()->like($column, $qb->expr()->literal('%' . $search . '%')));
```
Neloop: `search` parameter is not splitted, this causes that queries like `Martin Polanka` do not find anything.

app/helpers/Pagination.php
```diff
@@ -83,4 +111,50 @@ public function getOriginalOrderBy(): ?string {
...
+  /**
+   * Find appropriate MySQL Collation for given locale.
+   */
+  public function getLocaleCollation()
+  {
+    static $collations = null;
+    if (!$collations) {
+      // First time initialization ...
+      $collations = [
+        'cs' => 'utf8_czech_ci'
```
Neloop: What about `en` collation? Do you assume that it is default and therefore you do not have to set it?

app/V1Module/presenters/PipelinesPresenter.php
```diff
@@ -108,26 +108,31 @@ public function actionGetDefaultBoxes() {
...
+  public function actionDefault(int $offset = 0, int $limit = null, string $orderBy = null, array $filters = null, string $locale = null) {
+    $pagination = $this->getPagination($offset, $limit, $locale, $orderBy,
+      ($filters === null) ? [] : $filters, ['search', 'exerciseId', 'authorId']);
+
+    $pipelines = $this->pipelines->getPreparedForPagination($pagination);
     $pipelines = array_filter($pipelines, function (Pipeline $pipeline) {
       return $this->pipelineAcl->canViewDetail($pipeline);
     });
-    $this->sendPaginationSuccessResponse($pipelines, $pagination, []);
+    $this->sendPaginationSuccessResponse($pipelines, $pagination, true); // true = slice manually here
```
Neloop: more like automatically... no? :-)

app/model/repository/Pipelines.php
```diff
@@ -17,14 +21,46 @@ public function __construct(EntityManager $em) {
...
+      $search = trim($pagination->getFilter("search"));
+      if (!$search) {
+        throw new InvalidArgumentException("filter", "search query value is empty");
+      }
+
+      $qb->andWhere($qb->expr()->like("p.name", $qb->expr()->literal('%' . $search . '%')));
+    }
+
+    // Add order by
+    if ($pagination->getOrderBy() && !empty(self::$knownOrderBy[$pagination->getOrderBy()])) {
```
Neloop: If I am not mistaken, this piece of code is the same for `Users`, maybe a helper function would be on point? ... the same goes for detection of search filter?

app/V1Module/presenters/base/BasePresenter.php
```diff
@@ -245,22 +245,38 @@ protected function sendSuccessResponse($payload, $code = IResponse::S200_OK) {
...
-                                                   array $filters = [],
+                                                   bool $sliceItems = false,
+                                                   int $totalCount = null,
                                                    $code = IResponse::S200_OK) {
     $this->sendSuccessResponse([
-      "items" => array_slice(array_values($items), $pagination->getOffset(), $pagination->getLimit()),
-      "totalCount" => count($items),
+      "items" => $sliceItems
+        ? array_slice(array_values($items), $pagination->getOffset(),
+            $pagination->getLimit() ? $pagination->getLimit() : null)
```
Neloop: This should omit the zero? Or why is this ternary here?

app/model/repository/Exercises.php
```diff
@@ -47,22 +59,139 @@ public function searchByName(?string $search): array {
     }
 
     return $this->searchHelper($search, function ($search) {
-      $idsQueryBuilder = $this->em->createQueryBuilder()->addSelect("l.id")->from(LocalizedExercise::class, "l");
```
Neloop: The `searchByName` function is used somewhere else than in the `Exercises:default` action? Or why it stayed here?

app/V1Module/presenters/ExercisesPresenter.php
```diff
@@ -103,20 +109,52 @@ public function checkDefault() {
...
+    }
+  }
+
+  /**
+   * List authors of all exercises, possibly filtered by a group in which the exercises appear.
+   * @GET
+   * @param string $instanceId Id of an instance from which the authors are listed.
+   * @param string|null $groupId A group where the relevant exercises can be seen (assigned).
+   */
+  public function actionAuthors(string $instanceId = null, string $groupId = null)
```
Neloop: What is a purpose of this endpoint? And why don't you use mass endpoint to obtain users based on IDs?

tests/base/PresenterTestHelper.php
```diff
@@ -154,4 +154,27 @@ public static function getUser(Container $container, $login = null): User {
...
+   * @param $presenter The presenter which should handle the request.
+   * @param $module String representing the module path (e.g., 'V1:Exercises').
+   * @param $method HTTP method of the request (GET, POST, ...).
+   * @param $params Parameters of the request.
+   * @param $expectedCode Expected HTTP response code (200 by default).
+   * @return array|null Payload subtree of JSON request.
+   * @throws Tester\AssertException
+   * @throws JsonException
+   */
+  public static function preformPresenterRequest($presenter, string $module, string $method = 'GET', array $params = [], $expectedCode = 200)
```
Neloop: Typo... `preform`

app/helpers/Pagination.php
```diff
@@ -83,4 +111,50 @@ public function getOriginalOrderBy(): ?string {
...
+
+  public function getLocale()
+  {
+    return $this->locale;
+  }
+
+  /**
+   * Find appropriate MySQL Collation for given locale.
+   */
+  public function getLocaleCollation()
```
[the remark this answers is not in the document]
krulis-martin: I am not happy about it either. But there are very few options. Only other two choices would be to embed it in OrderByCollationInjectionMysqlWalker (which is slightly worse in my opinion) or place it in a separate helper (which is overkill in case of such a tiny function). If you have other alternatives in mind which I have overlooked, feel free to suggest them... :)

app/helpers/Pagination.php
krulis-martin: Exactly. But I can add it, if you would like.

app/model/repository/Exercises.php
krulis-martin: It is definitely used in tests since it is a convenient way to get a user. I have decided to keep it rather than fixing all the test (plus using pagination just to fetch one user by the name is a bit tedious).

app/model/repository/Pipelines.php
krulis-martin: OK. I will give it a serious thought.

app/model/repository/Users.php
krulis-martin: OK, can do.

app/helpers/Pagination.php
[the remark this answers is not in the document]
Neloop: For me, the better place would be the `OrderByCollationInjectionMysqlWalker`, but I do not insist on it, if you think it would be worse...

app/helpers/Pagination.php
Neloop: No, need... just to be sure...

app/V1Module/presenters/PipelinesPresenter.php
Neloop: yes, it would definitely be more clear ;-)

app/model/repository/Exercises.php
Neloop: Ok...

PR #226, opened 2018-07-27
Many2many exercise pipelines

krulis-martin:


app/model/entity/Pipeline.php
```diff
@@ -189,10 +205,10 @@ public function getHashedSupplementaryFiles(): array {
   /**
    * Create empty pipeline entity.
    * @param User $user
-   * @param Exercise|null $exercise
+   * @param Collection $exercises
    * @return Pipeline
    */
-  public static function create(User $user, Exercise $exercise = null): Pipeline {
+  public static function create(User $user, Collection $exercises = null): Pipeline {
```
Neloop: Default value null? It does not seem right.

app/model/entity/Pipeline.php
```diff
@@ -213,7 +229,7 @@ public static function create(User $user, Exercise $exercise = null): Pipeline {
    * @return Pipeline
    */
   public static function forkFrom(User $user, Pipeline $pipeline,
-      ?Exercise $exercise): Pipeline {
+    Collection $exercises = null): Pipeline {
```
Neloop: Again null to Collection...

app/V1Module/presenters/PipelinesPresenter.php
```diff
@@ -148,15 +148,22 @@ public function actionCreatePipeline() {
...
 
     // create pipeline entity, persist it and return it
-    $pipeline = Pipeline::create($this->getCurrentUser(), $exercise);
+    $pipeline = Pipeline::create($this->getCurrentUser());
     $pipeline->setName("Pipeline by {$this->getCurrentUser()->getName()}");
     $this->pipelines->persist($pipeline);
+    if ($exercise) {
+      $exercise->addPipeline($pipeline);
+      $this->exercises->persist($exercise);
+      $this->pipelines->refresh($pipeline); // this is needed so pipeline contains the newly added exercise
```
Neloop: You maybe should keep the original signature of `create` method on `Pipeline` and handing over only single `Exercise` and construct the `Pipeline` with proper `Collection` and not null. Therefore this whole `if` would not be necessary. The same approach was used in `Exercise` <-> `Group` relation if I am not mistaken.

app/V1Module/presenters/PipelinesPresenter.php
```diff
@@ -175,14 +182,21 @@ public function actionForkPipeline(string $id) {
...
+    }
 
     // fork pipeline entity, persist it and return it
-    $pipeline = Pipeline::forkFrom($this->getCurrentUser(), $pipeline, $exercise);
+    $pipeline = Pipeline::forkFrom($this->getCurrentUser(), $pipeline);
     $this->pipelines->persist($pipeline);
+    if ($exercise) {
+      $exercise->addPipeline($pipeline);
+      $this->exercises->persist($exercise);
+      $this->pipelines->refresh($pipeline); // this is needed so pipeline contains the newly added exercise
```
Neloop: Same as for `create` action.

app/model/repository/BaseRepository.php
```diff
@@ -72,6 +72,11 @@ public function flush() {
     $this->em->flush();
   }
 
+  public function refresh($entity)
+  {
```
Neloop: Bad format.

PR #227, opened 2018-07-27
New roles added and set role endpoint properly registered.

krulis-martin:


app/config/permissions.neon
```diff
@@ -168,7 +175,7 @@ permissions:
           - viewGroups
 
     - allow: true
-      role: supervisor
+      role: supervisor-student
```
Teyras: I don't think a supervisor-student should do anything with instances. Any counterexamples?

app/config/permissions.neon
```diff
@@ -130,7 +137,7 @@ permissions:
 
     - allow: true
       resource: group
-      role: supervisor
+      role: supervisor-student
       actions:
           - supervise
```
Teyras: This is a really bad privilege name. What can a user with the "supervise" privilege do?

app/V1Module/router/RouterFactory.php
```diff
@@ -331,6 +331,7 @@ private static function createUsersRoutes(string $prefix): RouteList {
     $router[] = new PostRoute("$prefix/<id>", "Users:updateProfile");
     $router[] = new PostRoute("$prefix/<id>/settings", "Users:updateSettings");
     $router[] = new PostRoute("$prefix/<id>/create-local", "Users:createLocalAccount");
+    $router[] = new PostRoute("$prefix/<id>/role", "Users:setRole");
```
Teyras: So this was implemented before and we didn't make a route for it?

krulis-martin: AFAIK, yes.

Neloop: Yes... this is my fault, I forgot the route.

app/config/permissions.neon
krulis-martin: Well, this is not mine and I did not change it. The meaning is that the user may be assigned as a supervisor of a group. I can rename it to "superviseGroup", or if you have better suggestions ...

Teyras: I know, it's even possible that I wrote it :) How about "becomeSupervisor"?

app/V1Module/router/RouterFactory.php
Teyras: No hard feelings, just curiosity :)

PR #229, opened 2018-07-31
Shadow assignments

Neloop:


app/V1Module/presenters/ShadowAssignmentsPresenter.php
```diff
@@ -0,0 +1,377 @@
...
+   * @throws BadRequestException
+   * @throws NotFoundException
+   */
+  public function actionCreate() {
+    $req = $this->getRequest();
+    $group = $this->groups->findOrThrow($req->getPost("groupId"));
+
+    if (!$this->groupAcl->canCreateShadowAssignment($group)) {
+      throw new ForbiddenRequestException("You are not allowed to create assignment in given group.");
+    }
```
krulis-martin: This should go to checkCreate().

app/V1Module/presenters/ShadowAssignmentsPresenter.php
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  /**
+   * Get a list of evaluations of all users for the assignment
+   * @GET
+   * @param string $id Identifier of the assignment
+   * @throws NotFoundException
+   */
+  public function actionEvaluations(string $id) {
```
Teyras: What on earth does evaluation mean in this context? Have you considered naming it something like PointAward?

app/V1Module/security/Policies/ShadowAssignmentEvaluationPermissionPolicy.php
```diff
@@ -0,0 +1,33 @@
...
+    $user = $identity->getUserData();
+
+    if ($user === null) {
+      return false;
+    }
+
+    return $group->isSupervisorOf($user) || $group->isAdminOf($user);
+  }
+
+  public function isEvaluatee(Identity $identity, ShadowAssignmentEvaluation $evaluation) {
```
Teyras: Wow, a new word!

app/model/entity/Group.php
```diff
@@ -413,12 +415,43 @@ public function getAssignments() {
...
+  }
+
   public function getMaxPoints(): int {
-    return array_reduce(
+    $points = array_reduce(
       $this->getAssignments()->getValues(),
       function ($carry, Assignment $assignment) { return $carry + $assignment->getGroupPoints(); },
       0
     );
+    $points = array_reduce(
```
Teyras: I would probably create two variables and add them up in the return statement here.

app/model/entity/ShadowAssignment.php
```diff
@@ -0,0 +1,157 @@
...
+    $this->shadowAssignmentEvaluations = new ArrayCollection;
+    $this->isPublic = $isPublic;
+    $this->localizedTexts = new ArrayCollection();
+    $this->version = 1;
+    $this->isBonus = $isBonus;
+    $this->createdAt = new \DateTime;
+    $this->updatedAt = new \DateTime;
+  }
+
+  public static function assignToGroup(Group $group, $isPublic = false) {
```
Teyras: I'm not sure I like this name - it kinda makes sense for exercises, but here it's just a weird constructor. Couldn't we just add `$group->addShadowAssignment()` to the constructor and remove this method alltogether?

app/model/entity/ShadowAssignment.php
```diff
@@ -0,0 +1,157 @@
...
+  protected $maxPoints;
+
+  /**
+   * Assignment can be marked as bonus, then we do not want to add its points
+   * to overall maximum points of group. This function will return 0 if
+   * assignment is marked as bonus one, otherwise it will return result of
+   * $this->getMaxPoints() function.
+   * @return int
+   */
+  public function getGroupPoints(): int {
```
Teyras: This is duplicated in `Assignment` - couldn't we put this somewhere where both classes can access it?

app/V1Module/security/Policies/ShadowAssignmentEvaluationPermissionPolicy.php
krulis-martin: https://en.wiktionary.org/wiki/evaluatee

app/V1Module/presenters/ShadowAssignmentsPresenter.php
```diff
@@ -0,0 +1,377 @@
...
+  public function actionCreate() {
+    $req = $this->getRequest();
+    $group = $this->groups->findOrThrow($req->getPost("groupId"));
+
+    if (!$this->groupAcl->canCreateShadowAssignment($group)) {
+      throw new ForbiddenRequestException("You are not allowed to create assignment in given group.");
+    }
+
+    if ($group->isOrganizational()) {
+      throw new BadRequestException("You cannot create assignment in organizational groups");
```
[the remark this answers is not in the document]
Neloop: Copied from the assignment, where you put it using this way...

app/V1Module/security/Policies/ShadowAssignmentEvaluationPermissionPolicy.php
Teyras: I know, I just didn't know it existed.

app/model/entity/ShadowAssignment.php
Neloop: Yeah, I though about having a base class and eventually end up with at least interface... but I can make base class for assignments...

Teyras: Or a trait...

app/V1Module/presenters/ShadowAssignmentsPresenter.php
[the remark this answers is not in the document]
Teyras: It's possible it was a mistake there - I had to change quite a bunch of endpoints.

app/model/entity/ShadowAssignment.php
Neloop: I would rather do a base class then a trait in this case.

app/V1Module/presenters/ShadowAssignmentsPresenter.php
Neloop: I thought, that @Teyras wanted to have only checks which uses query parameters in `check` methods, `groupId` is not a query param here.

app/V1Module/presenters/ShadowAssignmentsPresenter.php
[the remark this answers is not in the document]
Neloop: Ehm... The check method is missing here, the `groupId` is post parameter, according to your policy, this does not belong to check method.

app/model/entity/ShadowAssignment.php
Neloop: There are some default values in this factory function handed over to constructor. I would preserve this function and name it more for your likeliness... `createInGroup` maybe?

app/model/entity/Group.php
Neloop: Will do

app/model/entity/ShadowAssignment.php
Teyras: Yeah, that's better.

app/model/entity/Pipeline.php
```diff
@@ -19,7 +19,7 @@
  * @method string getDescription()
  * @method User getAuthor()
  * @method PipelineConfig getPipelineConfig()
- * @method int getVersion()
+ * @method Exercise getExercise()
```
krulis-martin: Exercises are already in many-to-many relation. I do not think this will work.

Neloop: ¯\_(ツ)_/¯ And how am I suppose to know that :D

krulis-martin: You have reviewed and approved my PR which implemented it 6 days ago...

app/V1Module/presenters/ShadowAssignmentsPresenter.php
```diff
@@ -0,0 +1,377 @@
...
+
+      $localizedTexts[$lang] = $localized;
+    }
+
+    // make changes to database
+    Localizations::updateCollection($assignment->getLocalizedTexts(), $localizedTexts);
+
+    foreach ($assignment->getLocalizedTexts() as $localizedText) {
+      $this->shadowAssignments->persist($localizedText, false);
+    }
```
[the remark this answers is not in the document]
Neloop: Quoting @Teyras 
> Because Doctrine doesn't really care what you pass to the `persist` method. The call itself is to let Doctrine know of any new localized texts that might have been created by the previous line.

app/model/entity/ShadowAssignmentPoints.php
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * @ORM\ManyToOne(targetEntity="User")
+   */
+  protected $author;
+
+  /**
+   * @ORM\ManyToOne(targetEntity="User")
+   */
+  protected $awardee;
```
krulis-martin: I think you also need to create unique index on columns shadowAssignment and awardee.

app/model/repository/ShadowAssignmentPointsRepository.php
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Find best solutions of given assignments for user.
+   * @param ShadowAssignment[] $shadowAssignments
+   * @param User $user
+   * @return ShadowAssignmentPoints[]
+   */
+  public function findEvaluationsForAssignment(array $shadowAssignments, User $user): array {
+    return $this->findBy([
+      "author" => $user,
+      "shadowAssignment" => $shadowAssignments,
```
krulis-martin: Sorry, I do not see, how this should work. Why is $shadowAssignments an array, passed to ManyToOne property shadowAssignment? I was expecting just one shadow assignment.

app/model/repository/ShadowAssignmentPointsRepository.php
```diff
@@ -0,0 +1,32 @@
...
+    parent::__construct($em, ShadowAssignmentPoints::class);
+  }
+
+  /**
+   * Find best solutions of given assignments for user.
+   * @param ShadowAssignment[] $shadowAssignments
+   * @param User $user
+   * @return ShadowAssignmentPoints[]
+   */
+  public function findEvaluationsForAssignment(array $shadowAssignments, User $user): array {
```
krulis-martin: Evaluations in the name -> should be points. And it does not make any sense, since Assignment is in singular here, yet an array of assignments is given.

app/model/repository/ShadowAssignmentPointsRepository.php
Neloop: Effing shiet... no this will not work

app/model/entity/ShadowAssignmentPoints.php
Neloop: Yes you think right

app/model/repository/ShadowAssignmentPointsRepository.php
Neloop: Well... actually yes... this will work, if you pass array to `shadowAssignment`, it'll find you all points entities for user and given shadow assignments...

app/model/view/ShadowAssignmentPointsViewFactory.php
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  /**
+   * @param ShadowAssignmentPoints[] $pointsList
+   * @return array
+   */
+  public function getPointsList(array $pointsList) {
+    return array_map(function (ShadowAssignmentPoints $evaluation) {
+      return $this->getPoints($evaluation);
+    }, $pointsList);
```
krulis-martin: Last "evaluation" relics.

Neloop: That rascal

PR #230, opened 2018-08-02
Single local login

Neloop:


app/V1Module/presenters/RegistrationPresenter.php
```diff
@@ -103,7 +103,7 @@ public function actionCreateAccount() {
 
     // check if the email is free
     $email = trim($req->getPost("email"));
-    if ($this->users->getByEmail($email) !== null) {
+    if ($this->logins->getByUsername($email) !== null) {
```
krulis-martin: This is somewhat strange -- getByUsername($email). This method should have a better name, or an explanatory comment is required.

Neloop: I do not follow... this is classical repository method which finds a login based on given username... what do you find strange about it?

krulis-martin: When you say user name, first that comes to mind is firstName and lastName. getByLogin or getByLoginName would be much better.

Neloop: But the column is `username`, therefore, repository convention says to have this method name...

Neloop: And no... we are not gonna rename the database column...

krulis-martin: OK, can you at least leave a note (in a comment) that username is in fact login?

Neloop: Can do

PR #233, opened 2018-08-06
Fixing ACLs for users.

krulis-martin:


app/config/permissions.neon
```diff
@@ -114,15 +114,22 @@ permissions:
 
     - allow: true
       resource: group
+      role: supervisor-student
```
Neloop: Just to let you know... this does not really do anything... You were probably just clarifying the permissions?

krulis-martin: Just a bit of defensive programming. And yes, second purpose is clarification.

PR #234, opened 2018-08-07
Python runner

Neloop:


migrations/Version20180805075758.php
```diff
@@ -0,0 +1,102 @@
...
+    $this->connection->beginTransaction();
+    $this->updatePipelines();
+    $this->connection->commit();
+  }
+
+  /**
+   * @param Schema $schema
+   */
+  public function down(Schema $schema) {
+  }
```
krulis-martin: Throw irreversible migration error at least?

Neloop: Oi... sorry... forgot to do that

PR #236, opened 2018-08-10
Changing the judge log database column to text and cropping added just to be sure

krulis-martin:


app/model/entity/TestResult.php
```diff
@@ -147,7 +147,7 @@ public function __construct(
   protected $message;
 
   /**
-   * @ORM\Column(type="string", nullable=true)
+   * @ORM\Column(type="text", length=4100, nullable=true)
```
Neloop: Why 4100?

krulis-martin: 4096 (default log size) + epsilon (to be sure) rounded up
But it is true that we can rely on worker to send only reasonably small logs and set this much higher (just in case worker fails to crop the log).

PR #237, opened 2018-08-10
Adding endpoint for setting the isAllowed flag of a user

krulis-martin:


app/V1Module/presenters/UsersPresenter.php
```diff
@@ -548,4 +548,25 @@ public function actionInvalidateTokens(string $id) {
...
+   * Set "isAllowed" flag of the given user. The flag determines whether a user may perform any operation of the API.
+   * @POST
+   * @param string $id Identifier of the user
+   * @Param(type="post", name="isAllowed", validation="bool", description="Whether the user is allowed (active) or not.")
+   * @throws InvalidArgumentException
+   * @throws NotFoundException
+   */
+  public function actionSetAllowed(string $id) {
+    $user = $this->users->findOrThrow($id);
+    $user->setIsAllowed((bool)$this->getRequest()->getPost("isAllowed"));
```
Neloop: I am not really sure if this will work... Did you check it and performed real HTTP request against this endpoint?

krulis-martin: No, it is tested by unit test only.

Neloop: Well if I am not mistaken, the parameter from the post will always be string... therefore conversion to boolean of `false` will result in `true`.

PR #238, opened 2018-08-11
Java jars hotfix removal and permanent solution proposed

Neloop:


app/helpers/ExerciseConfig/Pipeline/Box/Boxes/base/JavaUtilsTrait.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace App\Helpers\ExerciseConfig\Pipeline\Box;
+
+use App\Helpers\ExerciseConfig\Variable;
+
+trait JavaUtilsTrait
+{
```
krulis-martin: To be honest, I do not think this is a good example of a trait. The method being injected is a pure function. Hence, it does not require to reside in a class even less in the Box. In truth, I think this should be a Helper class, unless there is another reason for creating a trait (e.g., some functions that require $this), in which case this member function could simply tag along. Furthermore, the function should be static.

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/base/JavaUtilsTrait.php
```diff
@@ -0,0 +1,24 @@
...
+  protected function constructClasspath(?Variable $jarFiles) {
+    if ($jarFiles && !$jarFiles->isEmpty()) {
+      $classpath = ".";
+      foreach ($jarFiles->getValueAsArray() as $jar) {
+        $classpath .= ":" . $jar;
+      }
```
krulis-martin: Ever heard of join/import function?

migrations/Version20180810134011.php
```diff
@@ -0,0 +1,163 @@
...
+        if ($box["type"] === self::$JAVAC_BOX) {
+          $changed = true;
+          $changedPipelines[] = $pipeline["id"];
+
+          $box["portsIn"][self::$JAR_FILES_PORT] = [
+            "type" => self::$FILES_TYPE,
+            "value" => self::$JAR_FILES_PORT
+          ];
+          // input jar files in compilation
+          $config["boxes"][] = [
```
krulis-martin: Modifying an array over which you iterate using foreach inside the loop is dangerous and very error prone. Even though it will work in this case, I think it should be refactored.

migrations/Version20180810134011.php
```diff
@@ -0,0 +1,163 @@
...
+          ];
+          // output jar files from compilation
+          $config["boxes"][] = [
+            "name" => self::$PASSED_JARS_BOX_NAME,
+            "type" => self::$FILES_OUT_TYPE,
+            "portsIn" => [
+              "output" => [
+                "type" => self::$FILES_TYPE,
+                "value" => self::$JAR_FILES_PORT
+              ]
```
krulis-martin: Sorry, a bit confused on those two boxes here. FILES_IN_TYPE defines portsOut and FILES_OUT_TYPE defines portsIn?

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/base/JavaUtilsTrait.php
Neloop: Yes and I do not like them :-)

migrations/Version20180810134011.php
```diff
@@ -0,0 +1,163 @@
...
+          ];
+          $config["variables"][] = [
+            "name" => self::$JAR_FILES_PORT,
+            "type" => self::$FILES_TYPE,
+            "value" => []
+          ];
+        }
+      }
+
+      if ($changed) {
```
krulis-martin: This flag is redundant. You can get the same bit of information as !empty($changedPipelines); (which is even more comprehensible IMHO).

Neloop: Noted... this is a relict from previous implementation which was changed.

app/helpers/ExerciseConfig/Pipeline/Box/Boxes/base/JavaUtilsTrait.php
krulis-martin: I do. And I like them even better since they much shorter to write, much less error prone, and much more efficient (since you copy the string with every iteration twice!).

migrations/Version20180810134011.php
Neloop: yes this is correct

migrations/Version20180810134011.php
Neloop: To quote you
> do migraci uz nikdo nikdy nepujde

So if it works... let it work...

krulis-martin: OK, Just be careful next time.

Neloop: Will do master

migrations/Version20180810134011.php
Neloop: Well looking on the implementation... the `changed` variable is actually needed here. There are two `if` branches, the `changedPipelines` is changed only in one of them.

PR #239, opened 2018-08-16
Email debug and bug fixes

krulis-martin:


app/helpers/Emails/EmailHelper/EmailHelper.php
```diff
@@ -62,6 +70,41 @@ public function sendFromDefault(array $to, string $subject, string $text, array
     return $this->send(null, $to, $subject, $text, $bcc);
   }
 
+
+  /**
+   * Save mail copy into a text file in archiving directory.
+   * @param Message $message Message to be serialized into the file.
+   * @param \Excecption $lastMailerException Exception thrown from the sender (if any). Its message is logged as well.
```
Neloop: typo `\Excecption`, maybe it is not bad idea to `use` it

app/config/config.local.neon.example
```diff
@@ -39,6 +39,8 @@ parameters:
     apiUrl: https://recodex.mff.cuni.cz:4000
     footerUrl: https://recodex.mff.cuni.cz
     from: "ReCodEx <noreply@example.com>"
+    #debugMode: true  # uncomment for debugging
```
Neloop: the comments in global config file and local are not the same... in this case they are important because debug mode is defined very differently from the regular use case...

PR #246, opened 2018-09-25
Grand cleanup in groups endpoints.

krulis-martin:
Instances groups endpoint and all groups endpoints removed, default groups endpoint updated to support filters.

### DIFF
--- a/app/model/repository/Groups.php
+++ b/app/model/repository/Groups.php
@@ -4,9 +4,12 @@ namespace App\Model\Repository;
 
 use App\Model\Entity\Instance;
 use App\Model\Entity\LocalizedGroup;
+use App\Model\Entity\Group;
+use App\Model\Entity\GroupMembership;
+use App\Model\Entity\User;
+use App\Model\Helpers\PaginationDbHelper;
 use Doctrine\Common\Collections\Criteria;
 use Kdyby\Doctrine\EntityManager;
-use App\Model\Entity\Group;
 
 /**
  * @method Group findOrThrow($id)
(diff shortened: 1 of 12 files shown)


app/model/repository/Groups.php
```diff
@@ -103,13 +106,86 @@ public function findByName(string $locale, string $name, Instance $instance, ?Gr
...
+   */
+  public function groupsAncestralClosure(array $groups)
+  {
+    $res = [];
+    foreach ($groups as $group) {
+      $groupId = $group->getId();
+      if (array_key_exists($groupId, $res)) continue;
+      $res[$groupId] = $group;
+      foreach ($group->getParentGroupsIds() as $id) {
+        if (array_key_exists($id, $res)) break;
```
Neloop: Where in the name of our lord and savior are braces?

app/model/repository/Groups.php
```diff
@@ -103,13 +106,86 @@ public function findByName(string $locale, string $name, Instance $instance, ?Gr
...
+   * Gets an initial set of groups and produces a set of groups which have the
+   * original set as a subset and every group has its parent in the set as well.
+   * @param Group[] $groups Initial set of groups
+   */
+  public function groupsAncestralClosure(array $groups)
+  {
+    $res = [];
+    foreach ($groups as $group) {
+      $groupId = $group->getId();
+      if (array_key_exists($groupId, $res)) continue;
```
Neloop: same as below

app/model/repository/Groups.php
krulis-martin: As a convention, I do not write them in case of trivial statements (like if-break, if-return).

Neloop: Yeah... as a convention in our api, we write them everywhere

app/model/repository/Groups.php
```diff
@@ -170,11 +170,14 @@ public function groupsAncestralClosure(array $groups)
     $res = [];
     foreach ($groups as $group) {
       $groupId = $group->getId();
-      if (array_key_exists($groupId, $res)) continue;
+      if (array_key_exists($groupId, $res)) { // @neloop made me do that
```
Neloop: you are god damn right

PR #253, opened 2018-10-11
Fix cpp compilation

krulis-martin:


app/helpers/ExerciseConfig/Pipeline/Box/Boxes/Python3CompilationBox.php
```diff
@@ -97,7 +97,7 @@ public function getDefaultName(): string {
   public function compile(CompilationParams $params): array {
     $task = $this->compileBaseTask($params);
     $task->setCommandBinary(self::$PYTHON3_BINARY);
-    $task->setCommandArguments(["-m", "compileall", "-b", "."]);
+    $task->setCommandArguments(["python3", "-m", "compileall", "-b", "."]);
```
Neloop: Can you make `python3` constant too, we might want to change it some time...

PR #255, opened 2018-10-14
Compilation optimisation

Neloop:


app/helpers/ExerciseConfig/Compilation/BoxesOptimizer.php
```diff
@@ -14,19 +17,255 @@
...
+        // base $child has to be compared against all other children
+        // if the children are equal, then they should be optimised
+        while(!empty($copyChildren)) {
+          $compared = array_shift($copyChildren);
+
+          // children can be optimized... so do it
+          if (self::canNodesBeOptimized($child, $compared)) {
+            // delete the children from the base children array
+            $children = array_filter($children, function ($n) use ($compared) {
+              return $n === $compared;
```
krulis-martin: I am not sure, but I think that this condition should be reversed.

Neloop: It looks like, that it should...

app/helpers/ExerciseConfig/Compilation/BoxesOptimizer.php
```diff
@@ -14,19 +17,255 @@
...
+        $rootNode->addChild($node);
+        $node->addParent($rootNode);
       }
     }
+
+    // The hell of optimisation awaits...
+    // Forth, and fear no darkness! Arise! Arise, Riders of ReCodEx! Spears
+    // shall be shaken, shields shall be splintered! A sword day... a red day...
+    // ere the sun rises!
+    // Optimize! Optimize! OPTIMIZE!
```
krulis-martin: I would suggest minor modifications: Riders => Coders, Spears => Mice, shields => keys, sword => code.

app/helpers/ExerciseConfig/Compilation/DirectoriesResolver.php
```diff
@@ -0,0 +1,301 @@
...
+        $currentName = $currentName ?? $this->findOptimizedNodesDirectory($current);
+        $this->processNode($current, $currentName, $directoriesNodes);
+      }
+
+      // add children of current node into stack, the first children deserves better heritage from its father
+      // (because its the favourite child!), therefore it inherits the privilege to use the name of the directory,
+      // other children are not so lucky
+      if (count($current->getChildren()) > 0) {
+        $stack[] = [current($current->getChildren()), $currentName];
+        foreach (array_slice($current->getChildren(), 1) as $child) {
```
krulis-martin: It would be much readable (IMHO), if you just save $current->getChildren(), get the first item using array_shift() and then avoid using array_slice().

Neloop: That should be possible and maybe more readable... I consider it in my correction commit.

app/helpers/ExerciseConfig/Compilation/DirectoriesResolver.php
```diff
@@ -0,0 +1,301 @@
...
+  public function resolve(RootedTree $tree, CompilationContext $context, CompilationParams $params): RootedTree {
+    // Let's break it down...
+    // DirectoriesResolver works in cooperation with BoxesOptimizer which optimizes the flow of the tasks and marks the
+    // nodes which were optimized (this effectively means settings the test-id to null). Directories resolver then
+    // goes through the tree and creates the directories needed for execution. If the test-id is set, it is easy and
+    // straightforward, if it is not set the directories have to be smartly named and generated.
+    // The algorithm follows... The tree is searched with depth-first approach. Every node is processed in
+    // the following way. If the node belongs to the test, the test identification is recorded and children of this node
+    // are processed. If the node was optimised (has null test-id) then it is needed further processing. We need to
+    // figure out the name of the directory which will be created for this optimized node and its sub-nodes. The name of
```
Teyras: Wall of text. Could you please break it into bullets?

PR #260, opened 2018-10-26
Comments notifications

Neloop:


app/V1Module/presenters/CommentsPresenter.php
```diff
@@ -97,10 +118,19 @@ public function actionAddComment(string $id) {
...
     $comment = Comment::createComment($thread, $user, $text, $isPrivate);
 
-    $this->comments->persistComment($comment);
-    $this->comments->persistThread($thread);
+    $this->comments->persist($comment, false);
+    $this->comments->persist($thread, false);
     $this->comments->flush();
 
+    // send email to all participants in comment thread
+    if ($solution = $this->assignmentSolutions->get($id)) {
```
krulis-martin: It would be nice somehow emphasize that `=` inside is actually an assignment. I usually apply additional () around the assignment expression.

app/helpers/Emails/NotificationsHelper/Comments/SolutionCommentsEmailsSender.php
```diff
@@ -0,0 +1,142 @@
...
+  private $referenceSolutionCommentPrefix;
+
+  /**
+   * Constructor.
+   * @param EmailHelper $emailHelper
+   * @param array $params
+   */
+  public function __construct(EmailHelper $emailHelper, array $params) {
+    $this->emailHelper = $emailHelper;
+    $this->sender = Arrays::get($params, ["emails", "from"], "noreply@recodex.cz");
```
krulis-martin: The recodex.cz is nonexistent domain. Should not it be prudent to replace recodex.cz with recodex.mff.cuni.cz (or recodex.localhost if you insist on being completely independent of the founding institution).

app/helpers/Emails/NotificationsHelper/Comments/SolutionCommentsEmailsSender.php
```diff
@@ -0,0 +1,142 @@
...
+      $recipients[] = $user->getEmail();
+    }
+
+    // make user emails unique, so the emails won't be multiplied
+    $recipients = array_unique($recipients);
+
+    // filter out the author of the comment, it is pointless to send email to that user
+    $recipients = array_filter($recipients, function (string $email) use ($comment) {
+      return $email !== $comment->getUser()->getEmail();
+    });
```
krulis-martin: I know that premature optimizations are sqrt() of all evil; however, in this case it would be much nicer to user emails as keys in $recepients array. The uniqueness will take care of itself and removing logged user's email would not require filtering the array. And the only drawback is that you need to use array_keys($recepients) in the send() method...

app/helpers/Emails/NotificationsHelper/Comments/SolutionCommentsEmailsSender.php
```diff
@@ -0,0 +1,142 @@
...
+   * Prepare and format body of the assignment solution comment.
+   * @param AssignmentSolution $solution
+   * @param Comment $comment
+   * @return string Formatted mail body to be sent
+   */
+  private function createAssignmentSolutionCommentBody(AssignmentSolution $solution, Comment $comment): string {
+    // render the HTML to string using Latte engine
+    $latte = new Latte\Engine();
+    return $latte->renderToString(__DIR__ . "/assignmentSolutionComment.latte", [
+      "assignment" => $solution->getAssignment()->getLocalizedTexts()->first()->getName(), // TODO
```
krulis-martin: Well, the template is in English, I would suggest to get English name if available, then resort to the first localization.

app/helpers/Emails/NotificationsHelper/Comments/SolutionCommentsEmailsSender.php
```diff
@@ -0,0 +1,142 @@
...
+   * Prepare and format body of the reference solution comment.
+   * @param ReferenceExerciseSolution $solution
+   * @param Comment $comment
+   * @return string Formatted mail body to be sent
+   */
+  private function createReferenceSolutionCommentBody(ReferenceExerciseSolution $solution, Comment $comment): string {
+    // render the HTML to string using Latte engine
+    $latte = new Latte\Engine();
+    return $latte->renderToString(__DIR__ . "/referenceSolutionComment.latte", [
+      "exercise" => $solution->getExercise()->getLocalizedTexts()->first()->getName(), // TODO
```
krulis-martin: Same here.

app/helpers/Emails/NotificationsHelper/Comments/assignmentSolutionComment.latte
```diff
@@ -0,0 +1,6 @@
+<h3>New comment at assignment '{$assignment}' submitted by '{$solutionAuthor}'</h3>
+<p>
+    New comment was added under the solution of assignment '{$assignment}'
+    submitted by '{$solutionAuthor}'. Comment was added by '{$author}' on
+    {$date|date:'j.n.Y H:i'}.
+</p>
```
krulis-martin: I think it would be very helpful if the text of the comment was included as well. In fact, the text is far more important then the exact date-time! It would also be swell, if URL pointing to the solution (where the conversation is) was included.

app/V1Module/presenters/CommentsPresenter.php
Neloop: Like `if (($solution = $this->assignmentSolutions->get($id))) {}` ?

app/helpers/Emails/NotificationsHelper/Comments/assignmentSolutionComment.latte
Teyras: Do we already have a config option to specify the address of the web app?

Neloop: well we need the route for assignment, which ideally should be also configurable...

PR #264, opened 2018-11-01
Email localization

Neloop:


app/V1Module/presenters/EmailsPresenter.php
```diff
@@ -78,15 +86,17 @@ public function checkSendToSupervisors() {
    */
   public function actionSendToSupervisors() {
     $supervisors = $this->users->findByRoles(Roles::SUPERVISOR_ROLE, Roles::SUPERADMIN_ROLE);
```
krulis-martin: This is not a complete list of supervisor roles.

app/V1Module/presenters/EmailsPresenter.php
```diff
@@ -104,15 +114,17 @@ public function checkSendToRegularUsers() {
    */
   public function actionSendToRegularUsers() {
     $users = $this->users->findByRoles(Roles::STUDENT_ROLE);
```
krulis-martin: This is not a complete list of student roles.

app/V1Module/presenters/EmailsPresenter.php
Neloop: True, EmailsPresenter was implemented before role addition and I forgot about that in this reimplementation...

Neloop: The question is how to solve this in more forward-compatible way...

app/V1Module/presenters/EmailsPresenter.php
```diff
@@ -153,19 +167,18 @@ public function actionSendToGroupMembers(string $groupId) {
...
     }
 
-    $emails = array_map(function (User $user) {
-      return $user->getEmail();
-    }, $users);
-
     // user requested copy of the email to his/hers email address
-    if ($toMe && !in_array($user->getEmail(), $emails)) {
-      $emails[] = $user->getEmail();
+    if ($toMe && !in_array($user, $users)) {
```
krulis-martin: I do not like using in_array on objects (actually, I do not like the function at all, since it requires O(N) time). Comparison of two object invokes == operator on each property, which may not even work since these objects are Doctrine entities if I am not mistaken (there is no guarantee, that they may not have some internal members not related to the user's data). I believe that the only correct way here is to compare user IDs only.

app/helpers/Emails/EmailHelper/email_cs.latte
```diff
@@ -0,0 +1,69 @@
...
+                <hr>
+                <p>
+                  Pokud nechcete dostávat tyto emaily, aktualizujte prosím svoje uživatelské nastavení.
+                </p>
+              </td>
+            </tr>
+            <tr style="background: #f8f8f8; margin: 0; padding: 0; text-align: left; overflow: hidden;">
+              <td style="margin: 0; padding: 20px;">
+                <p style="margin: 0; padding: 0;">
+                  Prgejte zodpovědně,<br>ReCodEx Team<br>
```
krulis-martin: Something a little less jovial?

Neloop: Suggestions?

app/helpers/Emails/EmailHelper/email_cs.latte
```diff
@@ -0,0 +1,69 @@
...
+                </a>
+              </td>
+            </tr>
+          </table>
+        </td>
+      </tr>
+    </table>
+  </div>
+</body>
+</html>
```
krulis-martin: If I understand correctly, the structure of this template is a copy of English template. Hence, if we need to change the appearance, we need to do it correctly in two places. Since the template holds only a few messages, I believe that there should be only one template, and localization should be handled outside of the template.

app/V1Module/presenters/EmailsPresenter.php
krulis-martin: I do not think you can (at least not easily). Just update the list.

app/helpers/Emails/EmailHelper/email_cs.latte
krulis-martin: "Programujte zodpovědně"?

app/helpers/Emails/EmailHelper/email_cs.latte
Neloop: I agree, but I would like to resolve this the same way as mail prefixes, but that is entirely new functionality and deserves its own PR.

Neloop:
Apart the `email_cs` removal, all other comments were addressed in last commit.

PR #272, opened 2018-12-14
Fixing transaction processing to avoid TX aborts caused by deadlocks.

krulis-martin:


app/commands/cleanup/CleanupExerciseConfigs.php
```diff
@@ -163,25 +164,39 @@ private function cleanupTests(DateTime $limit): int {
...
-      $deletedTestsCount = $this->cleanupTests($limit);
-      $this->exercises->commit();
+    $tryAgain = 5;
+    while ($tryAgain-- > 0) {
+      $exception = null;
+      $this->exercises->beginTransaction();
+      try {
+        $this->executeUnsafe($limit, $output);
+        break;
+      }
```
Neloop: Nitpick :-) newline again in try-catch

PR #280, opened 2018-12-31
Proposal: Scheduler

Neloop:
This proposal is an unfinished draft (RFC) of possible implementation of internal **api** scheduler. The purpose of this PR is to portray main features of the scheduler and introduce the overall scheduler solution. If accepted the implementation shall be finished.


Neloop:
Will not be merged...

Some code will be reused in planned async jobs feature

PR #291, opened 2019-01-25
Fix issues with correct assignment visibility.

krulis-martin:


app/model/entity/Group.php
```diff
@@ -514,7 +514,7 @@ public function getLocalizedTexts(): Collection {
    */
   public function getPublicAssignments() {
```
Neloop: This method is not used anywhere in code, I suggest to take it down...

PR #293, opened 2019-01-27
Fix exercise assignments

krulis-martin:


app/helpers/Emails/Notifications/AssignmentPoints/PointsChangedEmailsSender.php
```diff
@@ -53,6 +53,11 @@ public function __construct(EmailHelper $emailHelper, array $params) {
    * @throws InvalidStateException
    */
   public function solutionPointsUpdated(AssignmentSolution $solution): bool {
+    if ($solution->getAssignment()->getGroup() === null) {
```
krulis-martin: What if the assignment was deleted? Or why `getAssignment()` works differently to other getters of soft-deletable entities?

Neloop: I must have forgotten about this

PR #295, opened 2019-01-30
Improving UX for assignment creation and updates

krulis-martin:


krulis-martin:
This is object-oriented system... If the users deem that 100 is not enough, they will object.

PR #300, opened 2019-02-05
Adding configuration and implicit groups to internal registration.

krulis-martin:


app/config/config.neon
```diff
@@ -51,6 +51,10 @@ parameters:
   monitor:  # connection to monitor
     address: wss://recodex.mff.cuni.cz:4443/ws  # URI of monitor
 
+  registration:
+    enabled: false  # for security reasons, simple registration is disabled by default
```
Neloop: This option should definitely be present in `config.local.neon.example`, so it will be visible to new-coming admins.

app/config/config.neon
```diff
@@ -51,6 +51,10 @@ parameters:
   monitor:  # connection to monitor
     address: wss://recodex.mff.cuni.cz:4443/ws  # URI of monitor
 
+  registration:
```
Neloop: I would probably name this option something like `local-registration` to distinguish it from general mechanism of registration to which even external ones can belong to.

app/config/config.local.neon.example
```diff
@@ -26,6 +26,10 @@ parameters:
   monitor:
     address: wss://recodex.mff.cuni.cz:443/ws
 
+  localRegistration:
+    enabled: false  # set to true if you want to permit simple local registration
+    implicitGroupsIds: {}  # list of group IDs to which each newly registered user is added as a member
```
Neloop: I should be more specific... this options should not be here, only the `enabled` option

app/helpers/Config/RegistrationConfig.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace App\Helpers;
+use Nette;
+use Nette\Utils\Arrays;
+
+class RegistrationConfig
```
Neloop: Some description in form of doc-comment would be nice here...

PR #303, opened 2019-02-14
If reject message from broker contains reason, collect it and signal …

Neloop:
…it in the exception message

### DIFF
--- a/app/helpers/BrokerProxy.php
+++ b/app/helpers/BrokerProxy.php
@@ -114,7 +114,7 @@ class BrokerProxy {
     if ($response[0] === self::REJECTED_RESULT) {
       $rejectMessage = "The broker rejected our request";
       if (count($response) > 1) {
-        array_pop($response);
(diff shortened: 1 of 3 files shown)


app/helpers/BrokerProxy.php
```diff
@@ -102,12 +103,22 @@ public function startEvaluation(string $jobId, array $hardwareGroups, array $hea
...
       throw new SubmissionFailedException("Receiving response from the broker failed.");
     }
 
+    // just close the damn connection
     $queue->disconnect($this->brokerAddress);
+
+    if ($response[0] === self::REJECTED_RESULT) {
+      $rejectMessage = "The broker rejected our request";
+      if (count($response) > 1) {
+        array_pop($response);
```
krulis-martin: `array_shift`?

Neloop: 🤦‍♂️

PR #305, opened 2019-02-19
User active

krulis-martin:


app/commands/RemoveInactiveUsers.php
```diff
@@ -0,0 +1,83 @@
...
+    $this->anonymizationHelper = $anonymizationHelper;
+  }
+
+  protected function configure() {
+    $this->setName('users:remove-inactive')->setDescription('Remove users who has not been active for some time.');
+    $this->addOption('silent', null, InputOption::VALUE_NONE, 'Just do the job (no interactions, no reports)');
+  }
+
+  protected function execute(InputInterface $input, OutputInterface $output) {
+    if (!$this->inactivityThreshold) return;
```
Neloop: PSR-2! :-)

app/commands/RemoveInactiveUsers.php
```diff
@@ -0,0 +1,83 @@
...
+      if (!$helper->ask($input, $output, $question)) {
+        $output->writeln("Aborted.");
+        return;
+      }
+    }
+
+    // perform deletion
+    foreach ($usersToDelete as $user) {
+      $this->anonymizationHelper->prepareUserForSoftDelete($user);
+      $this->users->remove($user);
```
Neloop: Just curious, is there some method for entity deletion which handles array of entities?

app/config/config.neon
```diff
@@ -148,6 +148,13 @@ parameters:
       solutionPointsUpdatedPrefix: "Assignment Points Updated - "
       shadowPointsUpdatedPrefix: "Shadow Assignment Points Awarded - "
 
+  generalStatsNotifications:
+    period: "1 week"
+    emails:
+      from: %emails.from%
+      to: %failures.emails.to%
```
Neloop: This is rather unexpected dependency on other specific configuration parameter, I would suggest to establish new general parameter with name for example `emails.adminEmail` or similar and use this one in `failures` and in here.

app/config/config.neon
```diff
@@ -191,6 +198,11 @@ parameters:
     anonymizedName: "@@anonymized"  # String used as replacement for anonymized name a user
     deletedEmailSuffix: "@@deleted.recodex"  # Suffix string appended to an email address of a user, when account is deleted
 
+  removeInactiveUsers:
+    # How long the user has to be inactive to warant the removal (null = never remove students, 1 month is minimum).
+    # Please note that the length of the auth. token expiration should be considered (readonly tokens may expire after 1 year).
```
Neloop: 👍

app/helpers/Emails/Notifications/SubmissionEmailsSender.php
```diff
@@ -37,7 +37,7 @@ class SubmissionEmailsSender {
   public function __construct(EmailHelper $emailHelper, array $params) {
     $this->emailHelper = $emailHelper;
     $this->sender = Arrays::get($params, ["emails", "from"], "noreply@recodex.mff.cuni.cz");
-    $this->submissionEvaluatedPrefix = Arrays::get($params, ["emails", "submissionEvaluatedPrefix"], "Submission Evaluated - ");
+    $this->submissionEvaluatedPrefix = Arrays::get($params, ["emails", "subject"], "Submission Evaluated - ");
```
Neloop: This feels like unintended change and victim of find-all replace-all and before that copy&paste :-)

app/helpers/GeneralStatsHelper.php
```diff
@@ -0,0 +1,205 @@
...
+ */
+class GeneralStatsHelper {
+  use Nette\SmartObject;
+
+  /** @var string */
+  private $inactivityThreshold;
+
+  /**
+   * @var Users
+   * @inject
```
Neloop: Inject? Really?

app/helpers/GeneralStatsHelper.php
```diff
@@ -0,0 +1,205 @@
...
+   */
+  public $users;
```
Neloop: Public? Really?

app/model/repository/BaseRepository.php
```diff
@@ -80,6 +101,10 @@ public function matching(Criteria $params) {
     return $this->repository->matching($params);
   }
 
+  public function createQueryBuilder(string $alias, string $indexBy = null)
```
Neloop: Is visibility `public` really on point here?

app/model/repository/ReferenceSolutionSubmissions.php
```diff
@@ -20,4 +14,20 @@ public function __construct(EntityManager $em) {
...
   }
 
+  /**
+   * Find all submissions created in given time interval.
+   * @param DateTime|null $since Only submissions created after this date are returned.
+   * @param DateTime|null $until Only submissions created before this date are returned.
+   */
+  public function findByCreatedAt(?DateTime $since, ?DateTime $until)
+  {
+    $qb = $this->createQueryBuilder('s'); // takes care of softdelete cases
```
Neloop: You wrote similar function three times in this PR, maybe probably some helper function on `BaseRepository` can be introduced?

app/commands/RemoveInactiveUsers.php
krulis-martin: Yeah, we do not break lines at the beginning of function and class declarations (which is one of the best things in PSR, but you choose to ignore it), but let's badger poor ol' me for not spreading a simple guard over 3 lines.

app/commands/RemoveInactiveUsers.php
krulis-martin: I do not think so.

app/helpers/GeneralStatsHelper.php
krulis-martin: Just copypasting your code :)

Neloop: Where from? In case this was in some helper, I shall be called idiot then...

krulis-martin: OK, maybe its not your code. I shall try to find and fix it at source as well.

PR #310, opened 2019-02-28
Adding defaultPage value to user settings.

krulis-martin:


app/V1Module/presenters/UsersPresenter.php
```diff
@@ -350,7 +351,9 @@ public function actionUpdateSettings(string $id) {
     $openedSidebar = $req->getPost("openedSidebar") !== null
       ? filter_var($req->getPost("openedSidebar"), FILTER_VALIDATE_BOOLEAN)
       : $settings->getOpenedSidebar();
-    $defaultLanguage = $req->getPost("defaultLanguage") !== null ? $req->getPost("defaultLanguage") : $settings->getDefaultLanguage();
+    $defaultLanguage = trim($req->getPost("defaultLanguage") !== null ? $req->getPost("defaultLanguage") : $settings->getDefaultLanguage());
```
Neloop: Usage of a `trim` only in initial condition of tertiary operator is a bit strange. In addition result from trim is never `null`, so the whole tertiary will always perform the first statement.

krulis-martin: However, the trim() engulfs the entire tertiary statement, so it is applied on its result.

Neloop: Oh, I see... in that case it is rather confusing line of code for me

PR #317, opened 2019-03-26
Introduce instance view factory

Neloop:


app/model/view/InstanceViewFactory.php
```diff
@@ -0,0 +1,61 @@
...
+   * @return array
+   */
+  public function getInstance(Instance $instance): array {
+    /** @var LocalizedGroup $localizedRootGroup */
+    $localizedRootGroup = Localizations::getPrimaryLocalization($instance->getRootGroup()->getLocalizedTexts());
+
+    return [
+      "id" => $instance->getId(),
+      "name" => $localizedRootGroup ? $localizedRootGroup->getName() : "",  # BC
+      "description" => $localizedRootGroup ? $localizedRootGroup->getDescription() : "",  # BC
```
krulis-martin: Is there a reason, why is BC comment started with `#` instead of `//`?

Neloop: No there is not

PR #326, opened 2019-06-05
Introducing a new property of the user entity for custom structured UI data and an endpoint for setting them.

krulis-martin:


app/model/view/UserViewFactory.php
```diff
@@ -55,13 +55,17 @@ private function getExternalIds(User $user, bool $canViewAllExternalIds = false)
     return $user->getConsolidatedExternalLogins($filter);
   }
 
+  private function isUserLoggedInUser(User $user) {
```
Neloop: This seems like something which should be checked by our authorization implementation, is it possible there?

app/model/entity/User.php
```diff
@@ -397,4 +397,28 @@ public function getCreatedAt(): DateTime {
   public function updateLastAuthenticationAt() {
     $this->lastAuthenticationAt = new DateTime();
   }
+
+
+  /**
+   * Arbitrary user-specific structured data stored by the UI.
```
Neloop: JSON-structured?

PR #327, opened 2019-08-21
Localized email subjects

Neloop:


app/helpers/Emails/EmailMacros.php
```diff
@@ -0,0 +1,41 @@
...
+class EmailMacros extends MacroSet {
+
+  public static function install(Latte\Compiler $compiler): MacroSet {
+    $set = new static($compiler);
+    $set->addMacro('emailSubject', [$set, 'macroEmailSubject'], [$set, 'macroEndEmailSubject']);
+    return $set;
+  }
+
+  public function macroEmailSubject(MacroNode $node, Latte\PhpWriter $writer) {
+    return 'ob_start(function () {})';
```
krulis-martin: I did not gather, why is it necessary to pass on empty lambda function as a callback to ob_start().

Neloop: Descriptive comment added

app/helpers/Emails/Notifications/Assignments/newShadowAssignmentEmail_en.latte
```diff
@@ -1,3 +1,4 @@
+{emailSubject}New shadow assignment - {$assignment}{/emailSubject}
```
krulis-martin: All other subjects are capitalized.

Neloop: Looks like some kind of leftovers... fixed

app/helpers/Emails/EmailVerificationHelper/EmailVerificationHelper.php
```diff
@@ -136,33 +128,24 @@ private function sendEmail(User $user, string $token, bool $firstTime = false):
...
    * @param User $user
    * @param string $locale
    * @param string $token
    * @param bool $firstTime
-   * @return string
+   * @return string[]
    * @throws InvalidStateException
    */
-  private function createBody(User $user, string $locale, string $token, bool $firstTime): string {
+  private function createEmail(User $user, string $locale, string $token, bool $firstTime): array {
```
krulis-martin: The return type hint should be `EmailRenderResult`.

app/helpers/Emails/Notifications/AssignmentPoints/PointsChangedEmailsSender.php
```diff
@@ -127,32 +120,31 @@ public function shadowPointsUpdated(ShadowAssignmentPoints $points): bool {
     }
 
     $locale = $awardee->getSettings()->getDefaultLanguage();
-    $subject = $this->shadowPointsUpdatedPrefix .
-      EmailLocalizationHelper::getLocalization($locale, $points->getShadowAssignment()->getLocalizedTexts())->getName();
+    list($subject, $text) = $this->createShadowPointsUpdated($points, $locale);
```
krulis-martin: If the create function yielded `EmailRenderResult`, this is not the best way how to get the properties out (I do not even think this will work without explicit cast to array).

PR #331, opened 2019-09-17
Superadmin may change any user password without verifying old password.

krulis-martin:


app/V1Module/security/Roles.php
```diff
@@ -35,14 +35,14 @@ protected function addRole(string $role, array $parents) {
...
   public function isInRole(string $actualTestedRole, string $minimalRequestedRole): bool {
-    if ($actualTestedRole === self::SUPERADMIN_ROLE) {
-      return true;  // special case -- superadmin takes it all...
-    }
-
     if ($actualTestedRole === $minimalRequestedRole) {
       return true;
     }
 
+    if ($actualTestedRole === self::SUPERADMIN_ROLE && substr($minimalRequestedRole, 0, strlen('scope-')) !== 'scope-') {
```
Teyras: Uh... Could we use `Nette\Utils\Strings::startsWith` here? I mean, we already have that library.

krulis-martin: Sure, I forgot we had fancy stuff like Nette\Utils...

PR #344, opened 2019-12-23
Reviewed flag in assignment solutions

Neloop:


app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -264,7 +264,71 @@ public function actionSetBonusPoints(string $id) {
...
+    if (!array_key_exists($flag, $knownBoolFlags)) {
+      throw new BadRequestException("Trying to set unknown boolean flag '$flag' to the solution");
+    }
+
+    // handle given flag
+    $unique = $knownBoolFlags[$flag];
+    $value = filter_var($req->getPost("value"), FILTER_VALIDATE_BOOLEAN);
+    // handle unique flags
+    if ($unique && $value) {
+      // flag has to be set to false for all other submissions
```
krulis-martin: Maybe a little bit more accurate here. That is all *solutions* of a user to an assignment (not all as all-*all*).

app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -264,7 +264,71 @@ public function actionSetBonusPoints(string $id) {
     $this->sendSuccessResponse("OK");
   }
 
-  public function checkSetAcceptedSubmission(string $id) {
+  public function checkSetFlag(string $id) {
+    $solution = $this->assignmentSolutions->findOrThrow($id);
+    if (!$this->assignmentSolutionAcl->canSetFlag($solution)) {
```
krulis-martin: Just not sure if one canSetFlag will cut it. But maybe we can leave it for now...

app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -274,11 +338,11 @@ public function checkSetAcceptedSubmission(string $id) {
   /**
    * Set solution of student as accepted, this solution will be then presented as the best one.
    * @POST
-   * @param string $id identifier of the submission
+   * @param string $id identifier of the solution
    * @throws \Nette\Application\AbortException
    * @throws NotFoundException
    */
-  public function actionSetAcceptedSubmission(string $id) {
+  public function actionSetAccepted(string $id) {
```
krulis-martin: Maybe also denote it as deprecated?

app/V1Module/router/RouterFactory.php
```diff
@@ -301,8 +301,10 @@ private static function createAssignmentSolutionsRoutes(string $prefix): RouteLi
     $router[] = new DeleteRoute("$prefix/<id>", "AssignmentSolutions:deleteSolution");
     $router[] = new PostRoute("$prefix/<id>/bonus-points", "AssignmentSolutions:setBonusPoints");
     $router[] = new GetRoute("$prefix/<id>/evaluations", "AssignmentSolutions:evaluations");
-    $router[] = new PostRoute("$prefix/<id>/set-accepted", "AssignmentSolutions:setAcceptedSubmission");
-    $router[] = new DeleteRoute("$prefix/<id>/unset-accepted", "AssignmentSolutions:unsetAcceptedSubmission");
+    $router[] = new PostRoute("$prefix/<id>/set-flag/<flag>", "AssignmentSolutions:setFlag");
+    $router[] = new PostRoute("$prefix/<id>/set-accepted", "AssignmentSolutions:setAccepted");
+    $router[] = new PostRoute("$prefix/<id>/set-accepted", "AssignmentSolutions:setAccepted");
```
krulis-martin: I think there is a copy-paste error... `set-accepted` twice?

app/model/entity/base/FlagAccessor.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace App\Model\Entity;
+
+use Exception;
+use ReflectionClass;
+
+trait FlagAccessor {
```
krulis-martin: Hmm... nice trait! :)

app/V1Module/presenters/AssignmentSolutionsPresenter.php
Neloop: You mean in doc comment? Or do we have some other ways?

app/V1Module/presenters/AssignmentSolutionsPresenter.php
Neloop: If you have better solution... I am not really sure if I do...

app/V1Module/presenters/AssignmentSolutionsPresenter.php
Neloop: Will do

app/V1Module/router/RouterFactory.php
Neloop: Wow... not really sure how this happened :-)

app/model/entity/base/FlagAccessor.php
Neloop: Well it felt a bit off to just copy `getFlag` and `setFlag` methods from `UserSettings` to `AssignmentSolution` :-)

app/V1Module/presenters/AssignmentSolutionsPresenter.php
krulis-martin: Doc comment. We have already done this ... search @DEPRECATED.

app/V1Module/presenters/AssignmentSolutionsPresenter.php
krulis-martin: I have, but it is not necessary at the moment.

Neloop: Just because I am curious, what is better solution? If it is explainable in few words?

krulis-martin: Flag is passed down to check function (since it is in URL), so we can have an ACL policy method for each flag (i.e., canSetFlagAccepted, canSetFlagReviewed). The name of the method can be assembled as a string from flag name (capitalization + prefix) and subsequently called as `$this->assignmentSolutionAcl->$methodName($solution)`.

PR #345, opened 2019-12-24
All Yaml operations wrapped with internal wrapper that uses PECL package (preferably) and falls back to Symfony\Yaml if need be (since it is much slower).

krulis-martin:


app/config/config.neon
```diff
@@ -9,7 +9,6 @@ parameters:
 
   jobConfig:
     storageDir: %appDir%/../job_config
-    humanReadable: true
```
Neloop: Where did this option go?

app/exceptions/MalformedJobConfigException.php
```diff
@@ -18,14 +18,14 @@ class MalformedJobConfigException extends ApiException {
...
-   * @param ParseException $originalException Optional pointer to original exception of YAML parser.
+   * @param YamlException $originalException Optional pointer to original exception of YAML parser.
    *                                          Line numbers and snippets can be found there.
    * @param string $frontendErrorCode
    * @param null $frontendErrorParams
    */
   public function __construct(
     string $msg = 'Please contact your supervisor',
-    ParseException $originalException = null,
+    YamlException $originalException = null,
```
Neloop: Maybe add typehint to `getOriginalException` method...

app/helpers/Yaml.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace App\Helpers;
+
+use Symfony\Component\Yaml\Yaml as SymfonyYaml;
+use Symfony\Component\Yaml\Exception\RuntimeException;
+
+
+class YamlException extends \RuntimeException {}
```
Neloop: Really with backslash?

app/helpers/Yaml.php
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * A Yaml wrapper to simplify transition from Symfony/Yaml to PECL yaml package.
+ * Historical note: we have switched to PECL since it is several orders of magnitude faster
+ * and performance become an issue when processing MiB-sized backend result yaml files.
+ * 
+ * Fallback to Symfony/Yaml is provided when extra formattin options are required.
```
Neloop: Typo: formattin

In addition, not really sure if fallback when options are provided is ideal... it is a bit hidden and not expected...

app/config/config.neon
krulis-martin: It was used in a single place for an unimportant config of Yaml output. No longer necessary, since yaml_emit is normally human-readable.

PR #347, opened 2019-12-30
After submit review notification mails

Neloop:


app/helpers/Emails/Notifications/Submissions/SubmissionEmailsSender.php
```diff
@@ -69,31 +70,45 @@ public function submissionEvaluated(AssignmentSolutionSubmission $submission): b
       return false;
     }
 
-    // Handle evaluation notification...
+    // Handle evaluation notifications...
+    return $this->handleEvaluationNotifications($solution, $submission, $user) &&
+      $this->handleTeacherNotifications($solution, $submission, $user);
```
krulis-martin: AFAIK the PHP interpret uses short-circuit evaluation. It means that teacher notifications will not be sent if the handleEvaluationNotifications() method yields false. Is this really desired behavior?

app/helpers/Emails/Notifications/Submissions/SubmissionEmailsSender.php
```diff
@@ -69,31 +70,45 @@ public function submissionEvaluated(AssignmentSolutionSubmission $submission): b
...
     }
 
-    // Handle teacher notifications...
-    $best = $this->assignmentSolutions->findBestSolution($assignment, $user);
-    $recipients = $this->getGroupSupervisorsRecipients($assignment->getGroup());
+    return true;
+  }
+
+  private function handleTeacherNotifications(AssignmentSolution $solution, AssignmentSolutionSubmission $submission, User $user): bool {
+    $success = true;
```
krulis-martin: I think this $success thing is an obfuscation which made it more difficult for me to understand the code. Furthermore, if we want to short-circuit, we should use return instead since the code is in a function. If not, this should be written in a different manner, so that reviewed email is sent even if accepted mail fails.

app/model/entity/UserSettings.php
```diff
@@ -149,6 +155,7 @@ public function jsonSerialize() {
       "solutionCommentsEmails" => $this->solutionCommentsEmails,
       "pointsChangedEmails" => $this->pointsChangedEmails,
       "assignmentSubmitAfterAcceptedEmails" => $this->assignmentSubmitAfterAcceptedEmails,
+      "assignmentSubmitAfterReviewedEmails" => $this->assignmentSubmitAfterReviewedEmails
```
krulis-martin: Add trailing comma, plz.

app/helpers/Emails/Notifications/Submissions/SubmissionEmailsSender.php
Neloop: Well what should I say, I am an idiot... of course this is not desired...

app/model/entity/UserSettings.php
Neloop: your wish is my command

Neloop:
Comments hopefully resolved by last commit.

PR #354, opened 2020-03-01
Packaging

SemaiCZE:
The package needs two dependencies:
- `yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm`
- `yum install https://rpms.remirepo.net/enterprise/remi-release-7.rpm`

Config, migrations, cron, php-fpm, ... need to be set up separately.

### DIFF
--- a/recodex-api.spec
+++ b/recodex-api.spec
@@ -1,0 +1,108 @@
+%define name recodex-api
+%define short_name api
+%define version 1.13.0
+%define unmangled_version cd4ae43bba0d4d29b196e2a7bfb622bcded794cb
+%define release 1
+
+Summary: ReCodEx API component
+Name: %{name}
+Version: %{version}
+Release: %{release}
+License: MIT
+Group: Development/Libraries
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-buildroot
+Prefix: %{_prefix}
+Vendor: Petr Stefan <UNKNOWN>
+Url: https://github.com/ReCodEx/api
+Requires(post): systemd
+Requires(preun): systemd
+Requires(postun): systemd
+Requires: composer php php-fpm php-json php-mysqlnd php-ldap php-pecl-zip php-pecl-zmq php-xml
+
+#Source0: %{name}-%{unmangled_version}.tar.gz
+Source0: https://github.com/ReCodEx/%{short_name}/archive/%{unmangled_version}.tar.gz#/%{short_name}-%{unmangled_version}.tar.gz
+
+%global debug_package %{nil}
+
+%description
+REST API of ReCodEx programmer testing solution.
+
+%prep
+%setup -n %{short_name}-%{unmangled_version}
+
+%build
+# Nothing to do here
+
+%install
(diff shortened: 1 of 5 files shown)


recodex-api.spec
```diff
@@ -0,0 +1,100 @@
...
+License: MIT
+Group: Development/Libraries
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-buildroot
+Prefix: %{_prefix}
+Vendor: Petr Stefan <UNKNOWN>
+Url: https://github.com/ReCodEx/api
+Requires(post): systemd
+Requires(preun): systemd
+Requires(postun): systemd
+Requires: git php73-php php73-php-json php73-php-mysqlnd php73-php-ldap php73-php-pecl-zip php73-php-pecl-zmq php73-php-xml php73-php-intl php73-php-mbstring
```
krulis-martin, line 20: First of all, I am concerned that these PHP packages are present only in remi repo. Is there a way how to specify repo-dependencies?
Furthermore, I think we are missing `php73-php-pecl-yaml` package in this list.

recodex-api.spec
```diff
@@ -0,0 +1,100 @@
...
+%pre
+getent group recodex >/dev/null || groupadd -r recodex
+getent passwd recodex >/dev/null || useradd -r -g recodex -d %{_sysconfdir}/recodex -s /sbin/nologin -c "ReCodEx Code Examiner" recodex
+exit 0
+
+%preun
+
+%files
+%defattr(-,recodex,recodex)
+%dir /opt/recodex-api/log
```
krulis-martin, line 70: Some other components (broker, monitor, worker) place their logs in /var/log/recodex.... shouldn't we follow this pattern and create at least a directory symlink in somewere in /var/log?

recodex-api.spec
```diff
@@ -0,0 +1,100 @@
...
+/opt/recodex-api/composer.lock
+/opt/recodex-api/composer-stable.phar
+%attr(0755,recodex,recodex) /opt/recodex-api/cleaner
+/opt/recodex-api/app/.htaccess
+/opt/recodex-api/www/.htaccess
+/opt/recodex-api/www/.maintenance.php
+
+%config /opt/recodex-api/app/config/config.neon
+%config /opt/recodex-api/app/config/permissions.neon
+%config(noreplace) /opt/recodex-api/app/config/config.local.neon
```
krulis-martin, line 97: Shouldn't we move this into /etc?

recodex-api.spec
```diff
@@ -0,0 +1,100 @@
...
+exit 0
+
+%preun
+
+%files
+%defattr(-,recodex,recodex)
+%dir /opt/recodex-api/log
+%dir /opt/recodex-api/job_config
+%dir /opt/recodex-api/uploaded_data
+%dir /opt/recodex-api/temp
```
krulis-martin, line 73: Another issue I have encountered ... log, job_config, uploaded_data, and temp must be accessible by Apache for writing. However, making then owned by apache:recodex seems sufficient.

PR #359, opened 2020-04-16
Adding 'isDetaining' flag to groups, which prevents students from leaving the group on their own.

krulis-martin:


app/V1Module/presenters/GroupsPresenter.php
```diff
@@ -166,6 +166,7 @@ public function actionDefault(
      * @Param(type="post", name="externalId", required=false, description="An informative, human readable identifier of the group")
      * @Param(type="post", name="parentGroupId", validation="string:36", required=false, description="Identifier of the parent group (if none is given, a top-level group is created)")
      * @Param(type="post", name="publicStats", validation="bool", required=false, description="Should students be able to see each other's results?")
+     * @Param(type="post", name="detaining", validation="bool", description="Can students leave the group on their own?")
```
Neloop, line 169: This seems a bit reversed... If I answer true to your question than the group will not be detaining...

krulis-martin: Right...

app/config/permissions.neon
```diff
@@ -111,7 +111,16 @@ permissions:
       role: student
       resource: group
       actions:
-        - removeStudent # A student can leave any group
+        - removeStudent # A student can leave any group...
```
Neloop, line 114: The comment seems unnecessary anymore :-)

PR #360, opened 2020-04-23
Score config refactoring

krulis-martin:


app/V1Module/presenters/ExercisesConfigPresenter.php
```diff
@@ -684,28 +685,36 @@ public function checkSetScoreConfig(string $id)
...
 
         $req = $this->getRequest();
+        $calculatorName = $req->getPost("scoreCalculator");
         $config = $req->getPost("scoreConfig");
 
         // validate score configuration
-        $calculator = $this->calculators->getCalculator($exercise->getScoreCalculator());
-        $normalizedConfig = $calculator->validateAndNormalizeScore($config);  // throws if validation fails
+        $calculator = $this->calculators->getCalculator($calculatorName);
+        $normalizedConfig = $config !== null
```
Neloop, line 704: I think, that in this case, better approach would be to `validateAndNormalizeScore` even for nulls. In current implementation you can have `weighted` calculator config with `null` value, which should be ideally avoided.

In uniform `validateAndNormalizeScore` you can easily check that the config is null and in weighted that the value is not null. 

Additionally in `WeightedScoreCalculator.getTestWeights(string)` method is non-nullable `string` parameter. But this method is called from `WeightedScoreCalculator.isScoreConfigValid(?string)` which has nullable string parameter...

migrations/Version20200421000600.php
```diff
@@ -0,0 +1,115 @@
...
+use App\Helpers\Evaluation\WeightedScoreCalculator;
+
+/**
+ * Manual migration.
+ * Identifies weighted score configs which are in fact uniform and change them accordingly.
+ */
+final class Version20200421000600 extends AbstractMigration
+{
+    private const UNIFORM_ID = 'uniform'; // This should match UniformScoreCalculator::ID
+    private const WEIGHTED_ID = WeightedScoreCalculator::ID; // This should match WeightedScoreCalculator::ID (former SimpleScoreCalculator)
```
Neloop, line 20: A bit weird :-) Why `weighted` id is taken from the class and `uniform` not?

app/helpers/Evaluation/EvaluationStatus/IEvaluable.php
```diff
@@ -14,29 +14,29 @@ interface IEvaluable
      * Query if the evaluation is ready
      * @return boolean The result
      */
-    function hasEvaluation(): bool;
+    public function hasEvaluation(): bool;
```
Neloop, line 17: The `public` seems a bit redundant... is there any special reason for it?

app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -537,4 +538,30 @@ public function actionDownloadResultArchive(string $evaluationId)
...
+    }
+
+    /**
+     * Get score configuration associated with given evaluation
+     * @GET
+     * @param string $evaluationId Identifier of the submission
+     * @throws NotFoundException
+     * @throws InternalServerException
+     */
+    public function actionEvaluationScoreConfig(string $evaluationId)
```
Neloop, line 558: Probably copy&paste error, please rename `$evaluationId` to `$submissionId`

app/V1Module/presenters/ReferenceExerciseSolutionsPresenter.php
```diff
@@ -665,4 +665,30 @@ public function actionDownloadResultArchive(string $evaluationId)
...
+    }
+
+    /**
+     * Get score configuration associated with given evaluation
+     * @GET
+     * @param string $evaluationId identifier of the reference exercise evaluation
+     * @throws NotFoundException
+     * @throws InternalServerException
+     */
+    public function actionEvaluationScoreConfig(string $evaluationId)
```
Neloop, line 685: Samesies... `$evaluationId` -> `$submissionId`

app/helpers/Evaluation/EvaluationStatus/IEvaluable.php
krulis-martin: It is required by PSR-12 validator.

migrations/Version20200421000600.php
krulis-martin: That was an attempt to minimize dependencies on the rest of the code. WeightedScoreCalculator was necessary, so even the ID was taken from there. But I can fix it if you so desire...

Neloop: No... just curious why...

PR #361, opened 2020-05-13
Changing the API of score configs so that Yaml encoding is handled internally and the config is always treated as a structure.

krulis-martin:


app/helpers/Evaluation/ScoreCalculators/UniformScoreCalculator.php
```diff
@@ -20,14 +20,7 @@ public function getId(): string
...
     }
 
-    /**
-     * Function that computes the resulting score from simple YML config and test results score
-     * @param string|null $scoreConfig
-     * @param array $testResults array of TestResult entities indexed by test ids
-     * @return float Percentage of total points assigned to the solution
-     * @throws SubmissionEvaluationFailedException
-     */
-    public function computeScore(?string $scoreConfig, array $testResults): float
```
Neloop, line 30: Why was the doc comment removed?

Neloop: Oh... it is in interface, is it?

krulis-martin: Indeed :)

PR #363, opened 2020-06-09
Universal score config

krulis-martin:


app/helpers/Evaluation/ScoreCalculators/UniversalScoreCalculator/AstNode.php
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Perform internal validation of the node itself (not the subtree).
+     * This method stands aside, so it can be reused and the main interface
+     * (validate() method) does not have to be overriden in derived classes.
+     * @param array $testNames List of allowed test names (array of strings)
+     * @throws AstNodeException if something is not valid
+     */
+    protected function internalValidation(array $testNames = [])
+    {
+        // expected to be overriden in derived classes
```
Neloop, line 147: If it is expected to be overriden, why is it not abstract?

krulis-martin: Because it is expected but not enforced. Empty implementation is the desired default.

PR #368, opened 2020-11-21
Filesystem

krulis-martin:


app/V1Module/presenters/AssignmentsPresenter.php
```diff
@@ -698,7 +704,7 @@ public function actionDownloadBestSolutionsArchive(string $id)
 
             $studentDir = Strings::webalize("{$student->getLastName()}_{$student->getFirstName()}_{$student->getId()}");
             foreach ($solution->getSolution()->getFiles() as $file) {
-                $files[$file->getLocalFilePath()] = $studentDir . "/" . $file->getName();
+                $files[$studentDir . "/" . $file->getName()] = $file->getFile($this->fileStorage);
```
Neloop, line 707: > $file->getFile($this->fileStorage

this seems like a bit reversed logic

krulis-martin: Depends on point of view, I suppose. However, the file returned from storage is wrapped in ImmutableFile object (abstraction over files), so we have no choice. Furthermore, I see logic in having the presentation name as key and the actual file (object in this case) as value.

app/http/ZipFilesResponse.php
```diff
@@ -54,9 +49,18 @@ private function compress()
...
-            if ($zip->addFile($localPath, $name) !== true) {
-                throw new ApiException("Error while adding file to archive");
+        foreach ($this->files as $name => $file) {
+            if ($file instanceof IImmutableFile) {
+                try {
+                    $file->addToZip($zip, $name);
+                } catch (FileStorageException $e) {
+                    throw new ApiException("Error while adding file to archive");
+                }
+            } else {
```
Neloop, line 59: Is this else really needed, is there situation where this would be used?

krulis-martin: You are right, it is not used anywhere. But I was thinking a bit ahead trying to make it reusable...

Neloop (approved):

PR #373, opened 2020-11-25
Repair some of the new phpstan errors

Neloop:


app/V1Module/security/AccessManager.php
```diff
@@ -74,12 +73,6 @@ public function decodeToken($token): AccessToken
             throw new InvalidAccessTokenException($token);
         } catch (UnexpectedValueException $e) {
             throw new InvalidAccessTokenException($token);
-        } catch (ExpiredException $e) {
-            throw new InvalidAccessTokenException($token);
-        } catch (SignatureInvalidException $e) {
-            throw new ForbiddenRequestException();
```
Neloop, line 80: Damn... this is not correct

app/helpers/PermissionHints.php
```diff
@@ -19,15 +19,15 @@ public function __construct()
...
+     * @param object $aclModule An ACL module
+     * @param object $subject The resource checked for permissions
      * @return Generator
      */
     public static function generate($aclModule, $subject)
     {
         foreach (static::getAclMethods($aclModule) as $method) {
             $parameter = $method->getParameters()[0];
-            $className = $parameter->getClass() ? $parameter->getClass()->getName() : null;
+            $className = $parameter->getClass()->getName();
```
Neloop, line 30: This is probably also not correct...

app/V1Module/security/AccessManager.php
Neloop: Repaired

app/helpers/PermissionHints.php
Neloop: This is actually bad typehint by nette... changed the code back

app/V1Module/security/UserStorage.php
```diff
@@ -35,12 +35,13 @@ public function __construct(AccessManager $accessManager, Nette\Http\IRequest $h
     }
 
     /**
-     * @param bool
-     * @return void
+     * @param bool $state
+     * @return static
```
krulis-martin, line 39: static?

app/commands/GenerateSwagger.php
```diff
@@ -122,15 +122,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         if (!$apiRoutes) {
             $output->writeln("<error>No suitable routes found</error>");
-            return;
+            return 1;
```
krulis-martin, line 125: What is this new devilry?

Neloop: Implementation of `execute` method of `Command` should return integer.

See https://github.com/symfony/console/blob/5.x/Command/Command.php#L163

app/V1Module/security/UserStorage.php
Neloop: Changed to `self` which is more self-explaining

krulis-martin (approved):

PR #377, opened 2020-12-10
Nette upgrade

krulis-martin:


app/helpers/Validators.php
```diff
@@ -16,10 +16,10 @@ class Validators extends Utils\Validators
      */
     public static function init()
     {
-        static::$validators['timestamp'] = [self::class, 'isTimestamp'];
+        static::$validators['timestamp'] = [static::class, 'isTimestamp'];
```
krulis-martin, line 19: Why static?

Neloop: Excellent question

Neloop (approved):

PR #379, opened 2020-12-14
Switch from kdyby/doctrine to nettrine doctrine integration

Neloop:


app/config/config.neon
```diff
@@ -445,34 +418,52 @@ services:
...
+nettrine.annotations:
   ignore:
     - LoggedIn
     - POST
     - GET
     - PUT
     - DELETE
+
+nettrine.extensions.atlantic18:
+    softDeleteable: true
```
krulis-martin, line 459: Is this the same as soft-deletable plugin we have been using, or did your replace it alltogethrer?

Neloop: We used `gedmo/doctrine-extensions` https://github.com/Atlantic18/DoctrineExtensions  
Aand, we are still using it through `nettrine/extensions-atlantic18` which is nette integration  
Soft-deletable functionality/filter is part of the doctrine-extensions.

krulis-martin: I.C.

krulis-martin (approved):

PR #387, opened 2021-02-28
Fix: correctly return 401 on invalid token

Neloop:
Fix: nette/security update changed behavior of User security structure, mimic the previous behavior by calling user-storage directly

### DIFF
--- a/app/V1Module/presenters/base/ApiErrorPresenter.php
+++ b/app/V1Module/presenters/base/ApiErrorPresenter.php
@@ -111,8 +111,11 @@ class ApiErrorPresenter extends \App\Presenters\BasePresenter
         string $frontendErrorCode = FrontendErrorMappings::E500_000__INTERNAL_SERVER_ERROR,
         $frontendErrorParams = null
     ) {
-        // log the action done by the current user
-        if ($this->getUser()->isLoggedIn()) {
+        // calling user->isLoggedIn results in throwing exception in case of
+        // invalid token (after update to nette/security:v3.1), therefore we
+        // need to call our UserStorage directly
+        if ($this->getUser()->getStorage()->isAuthenticated()) {
+            // log the action done by the current user
             // determine the action name from the application request
             $req = $this->getRequest();
             $params = $req->getParameters();
--- a/app/Bootstrap.php
+++ b/app/Bootstrap.php
@@ -11,8 +11,8 @@ class Bootstrap
         $configurator = new Configurator;
 
         //$configurator->setDebugMode('secret@23.75.345.200'); // enable for your remote IP
-        $configurator->enableTracy(__DIR__ . '/../log');
         // $configurator->setDebugMode(false);
+        $configurator->enableTracy(__DIR__ . '/../log');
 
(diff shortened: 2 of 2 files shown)


app/V1Module/presenters/base/ApiErrorPresenter.php
```diff
@@ -112,7 +112,7 @@ protected function sendErrorResponse(
         $frontendErrorParams = null
     ) {
         // log the action done by the current user
-        if ($this->getUser()->isLoggedIn()) {
+        if ($this->getUser()->getStorage()->isAuthenticated()) {
```
krulis-martin, line 115: I do not see why this change is necessary.

Neloop: Ok, let me tell ya, where the problem was.

If the token of the user was invalid then 500 HTTP status code was returned, that is obvious and utter rubbish. Although the correct exception was thrown and logged, so I needed to find why this was happening. Further investigations revealed that the response to the client was not our JSON but rather default nette fallback error page. Something started to smell fishy here...

After that I looked a bit more closely to the exception file and more importantly the stacktrace. The stacktrace revealed some unsettling thing and it all started to make some sense. It turned out the exception was thrown directly out of `ApiErrorPresenter`, which caused default nette fallback error page to take place. But why now, this never occured in previous years of operation, why now was the behavior so different that it mattered?

The culprit could be found on `ApiErrorPresenter:115` and some might already suspect why. That's right we updated nette and the implementation of `nette/security` changed significantly. So let me show you what changed and why it matter...

First we need to look at the previous implementation (v2.4) of `User->isLoggedIn` method:

https://github.com/nette/security/blob/v2.4/src/Security/User.php#L118

As we can see, it calls directly `isAuthenticated` method on storage and proxies all calls there. Now let's see new (v3.1) implementation of `User->isLoggedIn`:

https://github.com/nette/security/blob/v3.1/src/Security/User.php#L155

The implementation changed! And changed significantly. Now the call is not directly proxied, but rather handled in the `User` structure itself. The problem is `getStoredData` calls [`UserStorage->getState`](https://github.com/ReCodEx/api/blob/master/app/V1Module/security/UserStorage.php#L69) which returns flag if user is authenticated but also the identity...

The call to `UserStorage->getIdentity` is the real culprit here. It tries to parse the JWT token and fails with 401 exception which is thrown in the error handler `ApiErrorPresenter` and therefore cannot be handled there. So my proposed solution to this problem is to "reintroduce" the old logic in `ApiErrorPresenter` by calling `UserStorage->isAuthenticated` method directly. It feels like okay-ish solution, we kinda bent security mechanisms in the nette in the first place, so using directly our `UserStorage` should be okay. Although I also think that our security should be revised according to `nette/security:v3.1`...

krulis-martin: Well... that bending need at least a comment in the code. If you just lookup both methods in Nette documentation, they "do the same thing".

Neloop: commented

krulis-martin (approved):

PR #392, opened 2021-06-10
Async ops

krulis-martin:


app/V1Module/router/RouterFactory.php
```diff
@@ -588,4 +590,14 @@ private static function createWorkerFilesRoutes(string $prefix): RouteList
         $router[] = new PutRoute("$prefix/result/<type>/<id>", "WorkerFiles:uploadResultsFile");
         return $router;
     }
+
+    private static function createAsyncJobsRoutes(string $prefix): RouteList
+    {
+        $router = new RouteList();
+        $router[] = new GetRoute("$prefix/<id>", "AsyncJobs:default");
+        $router[] = new GetRoute("$prefix", "AsyncJobs:list");
+        $router[] = new PostRoute("$prefix/abort/<id>", "AsyncJobs:abort");
```
Neloop, line 599: Isn't more RESTful to have `/async-jobs/<id>/abort`? and not `/async-jobs/abort/<id>`?

app/config/config.neon
```diff
@@ -17,26 +17,44 @@ parameters:
...
     name: "ReCodEx API"
     description: "ReCodEx Code Examiner - API Server"
     versionFormat: "{tag}"
 
+  async:
+    pollingInterval: 60 # seconds (you may set this to larger values if inotify wakeups are allowed)
+    retries: 3 # how many times each async job is retried when failing
+    # inotify can wake the async worker (immediately once an async opertion is issued)
+    inotify: false # set to true only if your system (and PHP) supports inotify (not available on Windows, extension required on Linux)
+    inotifyFile: %tempDir%/async-inotify-file # file used as inotify rod for triggering async worker on dispatch
```
Neloop, line 30: I think that it would be better to add magic word `recodex` somewhere to the path, so it's clear to what app this file belongs to...

Also what does rod mean in this case?

app/helpers/Emails/EmailLocalizationHelper.php
```diff
@@ -110,4 +110,74 @@ function (User $user) {
 
         return $result;
     }
+
+    /**
+     *
```
Neloop, line 115: Empty? So sad...

install/recodex-core.service
```diff
@@ -0,0 +1,24 @@
+# ReCodEx core-api systemd configuration file
```
Neloop, line 1: The naming of the file is really very confusing, I would expect that unit with name `recodex-core` runs the REST Api ... and it does not, which is ... well, confusing

install/recodex-core.service
```diff
@@ -0,0 +1,24 @@
...
+# The core-api runs under PHP server, this service executes the async operations
+#
+# This file should be placed in /lib/systemd/system/ directory
+
+[Unit]
+Description=Recodex Core API Async Handler
```
Neloop, line 7: Maybe follow convention and use `ReCodEx`? :D

app/async/Worker.php
```diff
@@ -0,0 +1,236 @@
...
+        $this->asyncJobs->persist($job);
+        $this->asyncJobs->flush();
+    }
+
+    /**
+     * Main method of the worker that actually processes the async job.
+     * This method is blockinig, once it terminates the worker process should also terminate.
+     * @param string $workerId identifier of the worker (should be gathered from the CLI arguments)
+     */
+    public function run(string $workerId = 'worker1')
```
Neloop, line 207: I don't think that this should be parameter with default value, it should definitely be provided by caller everytime, to avoid future mistakes.

app/V1Module/router/RouterFactory.php
krulis-martin: True.

app/config/config.neon
krulis-martin: The tempDir is not global tmp, but local temp in the application directory (same place where Nette caches are).

The file acts like a *lightning rod* -- collects the touches (lightnings) and conducts them to a safe place (select sync operation). Perhaps this metafore has gone too far this time.

Neloop: Oh, I did not realize that about tempDir.

install/recodex-core.service
krulis-martin: I am open to suggestions. The service itself is closely attached to the core functionality, although the core is run by php-fpm which stands as a separate service. I thought that introducing another name would be even more confusing.

app/helpers/SubmissionHelper/SubmissionHelper.php
```diff
@@ -55,44 +114,183 @@ private function internalSubmit(
...
+     * @return array tuple containing created entities [ AssignmentSolutionSubmission, JobConfig ]
+     * @throws ForbiddenRequestException
+     * @throws InvalidArgumentException
+     * @throws ParseException
+     * @throws Exception
+     */
+    public function submit(AssignmentSolution $solution, User $user, bool $isDebug = false): array
+    {
+        if ($solution->getId() === null) {
+            throw new InvalidArgumentException("The submission object is missing an id");
```
Neloop, line 163: submission -> solution

app/model/entity/AsyncJob.php
```diff
@@ -0,0 +1,271 @@
...
+    {
+        return $this->startedAt;
+    }
+
+    /**
+     * @ORM\Column(type="datetime", nullable=true)
+     * @var DateTime|null
+     * This is the time when the task was either processed, killed, or deferred as unsolvable.
+     */
+    protected $terminatedAt = null;
```
Neloop, line 101: I think, that finishedAt would probably be a bit better

app/model/entity/AsyncJob.php
```diff
@@ -0,0 +1,271 @@
...
+        } else {
+            $this->error .= "\n$error";
+        }
+    }
+
+    /**
+     * @ORM\ManyToOne(targetEntity="Assignment", cascade={"persist", "remove"})
+     * @var Assignment|null
+     */
+    protected $associatedAssignment = null;
```
Neloop, line 230: This feels like stating the obvious...

app/V1Module/presenters/AsyncJobsPresenter.php
```diff
@@ -0,0 +1,156 @@
...
+use App\Exceptions\NotFoundException;
+use App\Exceptions\ForbiddenRequestException;
+use App\Exceptions\BadRequestException;
+use Doctrine\Common\Collections\Criteria;
+use Exception;
+use DateTime;
+
+/**
+ * Endpoints used by workers to exchange files with core.
+ * These endpoints take over responsibilities of FileServer component when integrated file-storage is used.
```
Neloop, line 18: Weird description... copy&paste?

app/V1Module/presenters/AsyncJobsPresenter.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        if (!$this->asyncJobsAcl->canList()) {
+            throw new ForbiddenRequestException("You cannot list async jobs");
+        }
+    }
+
+    /**
+     * Retrieves details about async jobs that are either pending or were recently completed.
+     * @GET
+     * @param int|null $ageThreshold Maximal time since completion (in seconds), null = only pending operations
```
Neloop, line 64: If I am not mistaken, we have `@Param` annotations with validations

Neloop: My bad, we cannot do <0 validation in those

app/V1Module/presenters/AsyncJobsPresenter.php
```diff
@@ -0,0 +1,156 @@
...
+        try {
+            $asyncJob = $this->asyncJobs->findOrThrow($id);
+            if ($asyncJob->getStartedAt() === null && $asyncJob->getTerminatedAt() === null) {
+                // if the job has not been started yet, it can be aborted
+                $asyncJob->setTerminatedNow();
+                $asyncJob->appendError("ABORTED");
+                $this->asyncJobs->persist($asyncJob);
+                $this->asyncJobs->commit();
+            } else {
+                $this->asyncJobs->rollback();
```
Neloop, line 129: Why rollback only read?

app/V1Module/presenters/AsyncJobsPresenter.php
krulis-martin: Yup.

app/V1Module/presenters/AsyncJobsPresenter.php
krulis-martin: Since it might be faster for the database (this intel is not verified though).

app/model/entity/AsyncJob.php
krulis-martin: Because obvious is underrated...

Neloop (approved):

PR #396, opened 2021-07-20
Update dependencies after removing magic accessors

Neloop:
After this update we are ready for PHP 8 from the dependencies perspective, although some of our code will have to be changed once we start supporting PHP 8...

### DIFF
--- a/app/helpers/PermissionHints.php
+++ b/app/helpers/PermissionHints.php
@@ -7,6 +7,7 @@ use Nette\Reflection\ClassType;
 use Nette\Reflection\Method;
 use Nette\StaticClass;
 use Nette\Utils\Strings;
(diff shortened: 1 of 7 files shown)


app/helpers/PermissionHints.php
```diff
@@ -24,8 +25,8 @@ public static function generate($aclModule, $subject)
     {
         foreach (static::getAclMethods($aclModule) as $method) {
             $parameter = $method->getParameters()[0];
-            /** @var ?ClassType $classObj */
-            $classObj = $parameter->getClass();
```
Neloop, line 28: `getClass` is deprecated in PHP 8, I am leaving this as a first early bird here...

krulis-martin (approved):

PR #397, opened 2021-07-21
Implementing max. points linear interpolation between first and second deadline.

krulis-martin:


tests/Points/Assignment.phpt
```diff
@@ -0,0 +1,97 @@
...
+
+    public function testMaxPointsTwoDeadlines()
+    {
+        $assignment = $this->getAssignment(0, 42, 10, 5);
+        Assert::equal(42, $assignment->getMaxPoints(self::getRelDateTime(-1))); // before first deadline
+        Assert::equal(5, $assignment->getMaxPoints(self::getRelDateTime(2))); // after first, before second
+        Assert::equal(0, $assignment->getMaxPoints(self::getRelDateTime(15))); // after second deadline
+    }
+
+    public function testMaxPointsInterpolation1()
```
Neloop, line 80: Hm... and where is the second one? If this is the first one? :D

Neloop (approved):

tests/Points/Assignment.phpt
krulis-martin: To be added in the future. Maybe.

PR #399, opened 2021-08-31
Solution files

krulis-martin:


app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -520,6 +521,40 @@ public function actionDownloadSolutionArchive(string $id)
...
+     * Get the list of submitted files of the solution.
+     * @GET
+     * @param string $id of assignment solution
+     * @throws ForbiddenRequestException
+     * @throws NotFoundException
+     */
+    public function actionFiles(string $id)
+    {
+        $solution = $this->assignmentSolutions->findOrThrow($id)->getSolution();
+        $entryPoint = $solution->getSolutionParams()->getVariable(ExecutionBox::$ENTRY_POINT_KEY);
```
Neloop, line 542: This feels weird, touching like really internal parts from frontend layer of the application...

app/V1Module/presenters/AssignmentSolutionsPresenter.php
```diff
@@ -520,6 +521,40 @@ public function actionDownloadSolutionArchive(string $id)
...
+        $entryPoint = $entryPoint ? $entryPoint->getValue() : null;
+
+        $files = $solution->getFiles()->toArray();
+        foreach ($files as &$file) {
+            $file->prepareExtendedSerializationData($this->fileStorage);
+            $file = $file->jsonSerialize();
+            if ($file["name"] === $entryPoint) {
+                $file["isEntryPoint"] = true;
```
Neloop, line 550: This arbitrary change of the response JSON structure are probably hard to document, might be better to somehow avoid them as much as possible... Just a thought

app/model/entity/SolutionZipFile.php
```diff
@@ -35,4 +35,31 @@ public function getFile(FileStorageManager $manager): ?IImmutableFile
...
     }
+
+    public function getNestedFile(FileStorageManager $manager, string $entryName): ?IImmutableFile
+    {
+        return $manager->getSolutionFile($this->getSolution(), $entryName);
+    }
+
+    private $zipEntries = null;
+
+    public function prepareExtendedSerializationData(FileStorageManager $manager): void
```
Neloop, line 46: I believe we introduced concept of views, which constructs JSON responses, in this case it seems like it would be much better solution, than filling entity with response data...

app/V1Module/presenters/AssignmentSolutionsPresenter.php
krulis-martin: I thought about creating solution files view, but it seemed like an overkill at that moment. However, it is true that the code has perhaps overgrown the threshold and it would make sense now.

Neloop (approved):

PR #400, opened 2021-09-18
Creating new endpoint for retrieving all solutions of a user from all assignments in a group at once.

krulis-martin:


app/V1Module/presenters/GroupsPresenter.php
```diff
@@ -852,6 +875,47 @@ public function actionStudentsStats(string $id, string $userId)
...
+     * @param string $id Identifier of the group
+     * @param string $userId Identifier of the student
+     * @throws BadRequestException
+     */
+    public function actionStudentsSolutions(string $id, string $userId)
+    {
+        $user = $this->users->findOrThrow($userId);
+        $group = $this->groups->findOrThrow($id);
+
+        if ($group->isStudentOf($user) === false) {
```
Neloop, line 901: Isn't this checked in `check` method using ACL?

Neloop: Thinking about this... even if it is, having this check and returning 400 makes sense...

Neloop (approved):

PR #403, opened 2021-09-29
Bug fixes

krulis-martin:


app/helpers/FileStorageManager.php
```diff
@@ -167,9 +167,10 @@ public function partialFileChunksCleanup(array $ids): int
         }
 
         // filter the files using the ids index
-        return $this->fileStorage->deleteByFilter(function ($file) use ($idsIndex) {
+        $now = time();
+        return $this->fileStorage->deleteByFilter(self::PARTIAL_UPLOADS . '/', function ($file) use ($idsIndex, $now) {
             $id = substr($file->getName(), 0, 36); // get the uuid, which is the prefix of the file name
-            return array_key_exists($id, $idsIndex);
+            return array_key_exists($id, $idsIndex) || $file->getTime() + 3600 > $now;
```
Neloop, line 173: This `3600` magic constant means what exactly?

krulis-martin: The thing is we need some delay to avoid deleting fresh upload files which have their entities added into DB right after cleanup script reads them. The script should take a few seconds max. so 1 hour (3600s) should be enough as a margin. I do not feel the need to make this configurable.

Neloop: Ok, then maybe change in the method comment is needed

> All files except those with existing ID prefixes are deleted.

Neloop (approved):

PR #413, opened 2021-11-13
Updating ACLs to divide assignExercise permission between group and exercise, which can be checked independently.

krulis-martin:


app/config/permissions.neon
```diff
@@ -356,24 +356,6 @@ permissions:
...
-      conditions:
-          - exercise.isSubGroupSupervisor
-          - group.isNotArchived
-
-    - allow: true
-      resource: group
-      actions:
-          - assignExercise
-      conditions:
-          - exercise.isPublic
```
Neloop, line 376: Seems like this condition was lost... is that correct?

app/model/repository/Groups.php
```diff
@@ -144,6 +150,54 @@ private function fetchGroupsByMembership(User $user, bool $admins): array
...
+        if (!$parentsIds) {
+            return [];
+        }
+        $qb = $this->createQueryBuilder('g'); // takes care of softdelete cases
+        $qb->andWhere($qb->expr()->in("g.parentGroup", $parentsIds));
+        return $qb->getQuery()->getResult();
+    }
+
+    /**
+     * Find all groups in are supervised by a user (as supervisor, admin, or admin inherently).
```
Neloop, line 170: A bit weird formulation, `in` does not seem to fit there...

app/config/permissions.neon
krulis-martin: It is. New `exercise.assign` permission embeds this condition in `isSubGroupNonStudentMember` and `isSuperGroupAdmin`, although it is not visible in the permissions file.

Neloop (dismissed):

app/model/repository/Groups.php
krulis-martin: Fixed.

Neloop (approved):

PR #414, opened 2021-11-18
Exercises optimization

krulis-martin:


app/V1Module/security/Policies/ExercisePermissionPolicy.php
```diff
@@ -56,16 +65,31 @@ public function isSubGroupNonStudentMember(Identity $identity, Exercise $exercis
...
+        $subgroupCache = &$this->subgroupMembersCache[$user->getId()];
+
         /** @var Group $group */
         foreach ($exercise->getGroups() as $group) {
-            if ($group->isNonStudentMemberOfSubgroup($user)) {
-                return true;
+            if (!array_key_exists($group->getId(), $subgroupCache)) {
+                $subgroupCache[$group->getId()] = $group->isNonStudentMemberOfSubgroup($user);
             }
+            return $subgroupCache[$group->getId()];
```
Neloop, line 78: Isn't this logic a bit changed? The return is no longer in the IF, therefore the result of the first group in array is returned...

app/V1Module/security/Policies/ExercisePermissionPolicy.php
```diff
@@ -77,17 +101,23 @@ public function isSuperGroupAdmin(Identity $identity, Exercise $exercise)
...
+        $supergroupCache = &$this->supergroupAdminCache[$user->getId()];
+
         /** @var Group $group */
         foreach ($exercise->getGroups() as $group) {
-            if ($group->isAdminOf($user)) {
-                return true;
+            if (!array_key_exists($group->getId(), $supergroupCache)) {
+                $supergroupCache[$group->getId()] = $group->isAdminOf($user);
             }
+            return $supergroupCache[$group->getId()];
```
Neloop, line 114: The same as above... no longer in if, only first group is iterated...

app/V1Module/security/Policies/ExercisePermissionPolicy.php
krulis-martin: Right, there should have been another IF.

Neloop (approved):

PR #423, opened 2022-02-19
Php74

krulis-martin:


app/V1Module/security/AuthorizatorBuilder.php
```diff
@@ -40,14 +41,15 @@ public function build($aclInterfaces, $permissions, $uniqueId): ClassType
         $check->addParameter("privilege")->setType("string");
 
         foreach ($permissions as $i => $rule) {
-            $all = new PhpLiteral(sprintf("%s::ALL", Permission::class));
```
Neloop, line 43: Does this change really do the same thing as the original code?

krulis-martin: It should, since `::ALL` is not used anywhere and the condition was constructed so it always returned false.

Neloop (approved):

PR #435, opened 2022-10-04
Code reviews

krulis-martin:


app/V1Module/router/RouterFactory.php
```diff
@@ -377,6 +377,16 @@ private static function createAssignmentSolutionsRoutes(string $prefix): RouteLi
             "AssignmentSolutions:evaluationScoreConfig"
         );
 
+        $router[] = new GetRoute("$prefix/<id>/review", "AssignmentSolutionReviews:");
+        $router[] = new PostRoute("$prefix/<id>/review", "AssignmentSolutionReviews:update");
+        $router[] = new DeleteRoute("$prefix/<id>/review", "AssignmentSolutionReviews:remove");
+        $router[] = new PostRoute("$prefix/<id>/review-comment", "AssignmentSolutionReviews:newComment");
+        $router[] = new PostRoute("$prefix/<id>/review-comment/<commentId>", "AssignmentSolutionReviews:editComment");
```
Neloop, line 384: PUT?

app/commands/SendPendingReviewsNotification.php
```diff
@@ -0,0 +1,89 @@
...
+        parent::__construct();
+        $this->threshold = $threshold;
+        $this->sender = $sender;
+        $this->assignmentSolutions = $assignmentSolutions;
+    }
+
+    protected function configure()
+    {
+        $this->setName(self::$defaultName)->setDescription(
+            'Send notifications for pending (not closed) code revirews to group admins.'
```
Neloop, line 42: revirews -> reviews

app/commands/SendPendingReviewsNotification.php
```diff
@@ -0,0 +1,89 @@
...
+        );
+    }
+
+    private $groupAdminsCache = [];
+
+    private function getGroupAdmins(Group $group): array
+    {
+        if (!array_key_exists($group->getId(), $this->groupAdminsCache)) {
```
Neloop, line 50: This if is weird... there is no body!

app/model/entity/AssignmentSolution.php
```diff
@@ -66,9 +67,23 @@ public function getMaxPoints()
...
     /**
-     * @ORM\Column(type="boolean")
+     * @ORM\Column(type="datetime", nullable=true)
+     * Time when a review was started.
+     */
+    protected $reviewStartedAt = null;
+
+    /**
+     * @ORM\Column(type="datetime", nullable=true)
+     * Time when a review was closed (and comments made visible to student).
```
Neloop, line 77: I cannot find any logic related to `(and comments made visible to student).` where is this handled?

app/model/repository/BaseRepository.php
```diff
@@ -36,6 +36,11 @@ public function __construct(EntityManagerInterface $em, $entityType)
         $this->repository = $repository;
     }
 
+    public function getEntityManagerForDebugging(): EntityManagerInterface
```
Neloop, line 39: Weird method... should it stay?

app/commands/SendPendingReviewsNotification.php
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return $this->groupAdminsCache[$group->getId()];
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $threshold = new DateTime();
+        if ($this->threshold) {
+            $threshold->modify($this->threshold); // only reviews opened before treshold are reported
```
Neloop, line 59: treshold -> threshold

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
```diff
@@ -0,0 +1,407 @@
...
+
+        if ($solution->getReviewStartedAt() === null) {
+            $solution->setReviewStartedAt(new DateTime()); // the review is marked as opened in any case
+        }
+
+        if ($close && $reviewedAt === null) {
+            // opened -> closed transition
+            $solution->setReviewedAt(new DateTime());
+
+            $issues = 0; // issues are duly counter only when the review is closed
```
Neloop, line 126: Maybe add this comment also to the AssignmentSolutionEntity->issuesCount member?

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
```diff
@@ -0,0 +1,407 @@
...
+            $this->reviewsEmailSender->solutionReviewRemoved($solution, $closed);
+        }
+
+        $this->sendSuccessResponse("OK");
+    }
+
+    public function checkNewComment(string $id)
+    {
+        $solution = $this->assignmentSolutions->findOrThrow($id);
+        if (!$this->assignmentSolutionAcl->canReview($solution)) {
```
Neloop, line 206: Shouldn't it be `canAddReviewComment`?

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
```diff
@@ -0,0 +1,407 @@
...
+
+    public function checkEditComment(string $id, string $commentId)
+    {
+        $solution = $this->assignmentSolutions->findOrThrow($id);
+        $comment = $this->reviewComments->findOrThrow($commentId);
+        if (!$this->assignmentSolutionAcl->canEditReviewComment($solution, $comment)) {
+            throw new ForbiddenRequestException("You are not allowed to edit this review comment");
+        }
+
+        if ($comment->getSolution()->getId() !== $id) {
```
Neloop, line 291: Couldn't this be done in the permissions?

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
```diff
@@ -0,0 +1,407 @@
...
+
+        $solution = $this->assignmentSolutions->findOrThrow($id);
+        if ($solution->getReviewedAt() !== null) {
+            if ($isIssue) {
+                $solution->setIssuesCount($solution->getIssuesCount() - 1);
+                $this->assignmentSolutions->persist($solution);
+            }
+            $this->reviewComments->flush();
+
+            // review has been closed, report the modification
```
Neloop, line 382: Seems a bit like this comment does not belong...

Neloop (changes requested):
See comments

app/commands/SendPendingReviewsNotification.php
krulis-martin: Good catch, I forget to implement this.

app/model/entity/AssignmentSolution.php
krulis-martin: I will rephrase that. When the review is closed, a student is notified via email where all comments are sent.
The only thing that is a little weird here is that the "issuesCount" is updated only after the review is closed (so the students won't start panicking too early when the review is still not finished).

app/model/repository/BaseRepository.php
krulis-martin: Yes, it might be helpful for future debugging sessions...

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
krulis-martin: Yep.

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
krulis-martin: Technically yes, but I did not like it. This is not a matter of ACL, this is a matter of verifying endpoint inputs. The endpoint gets two IDs, where one is actually implied by the other. I did not want to change the endpoint (for reasons risen at the frontend), so verification must be performed (also note that I am throwing BadRequest, not Forbidden exception).

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
krulis-martin: Not really, but I have made it better

app/V1Module/router/RouterFactory.php
krulis-martin: I thought about it and I think arguments can be made both for POST and PUT (I assume we are talking about the new comment, not the update). If you have a good point about why this should be PUT and should not be POST, I am willing to change it :)

Neloop: I am looking at other endpoints and we are not using PUT anywhere (with 2 exceptions)... so it does not make sense to make this one PUT if we generally do not use it...

But I am talking about updating the comment, updates are usually done in REST API with PUT, POST is for creation...

app/V1Module/presenters/AssignmentSolutionReviewsPresenter.php
Neloop: Ok, makes sense

app/V1Module/router/RouterFactory.php
krulis-martin: True, but
- PUT's semantics is often perceived as "replace", which is not entirely true for this endpoint (although this is only a weak argument)
- unlike POST, PUT has no input limit in the PHP input wrapper, thus we use it only for file uploads
So I would prefer consistency over ideology in this case.

Neloop: Yes, as I said, I would too...

Neloop (approved):

PR #469, opened 2024-10-17
Swagger

eceltov:
Added a new command `generate-swagger` that prints a swagger document representing the current API to the console.


app/commands/GenerateSwagger.php
```diff
@@ -2,626 +2,38 @@
...
+            'Generate a swagger specification file from existing code.'
+        );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $path = __DIR__ . '/../V1Module/presenters/_autogenerated_annotations_temp.php';
+        $openapi = \OpenApi\Generator::scan([$path]);
+
+        header('Content-Type: application/x-yaml');
```
krulis-martin, line 31: What is the purpose of setting a header here?

